PR #34, opened 2020-10-21
Add is_a directed search functionality to term search

Freymaurer:
Features added:
- Add is_a directed search functionality to term search.
- Restored bulma imports in style sheets for bulma-extensions switch and tooltip.
- change addBuildingBlock function to add 3 columns with two cols hidden. These extra columns will hold additional ontology term info in the future
- add new sql dump with is_a directed search procedure and is_a relationships in TermRelationship table

Do not merge until:

- [x] is_a directed search returns all children and not only direct children of is_a relationships.
- [x] insert term function also inserts the additional info in the new extra columns of addBuildingBlock.


src/Client/Views/BaseView.fs
```diff
@@ -27,6 +27,7 @@ let createNavigationTab (pageLink: Routing.Page) (model:Model) (dispatch:Msg-> u
             ]
         ] [
             Text.span [] [
+                /// does not work for me in excel, and in excel online i have a fixed width that always shows the icon
```
kMutagene, line 30: Please open an issue if this does not work anymore

src/Server/Server.fs
```diff
@@ -94,10 +108,25 @@ let testWebApp =
     |> Remoting.buildHttpHandler
 
 
+let annotatorDocs = Docs.createFor<IAnnotatorAPI>()
```
kMutagene, line 111: Does this work well? if so, we should expand it and provide an api doc link in the repo and on the page

kMutagene (approved):
Good stuff. Please check my comments, but this can be merged as is.

src/Client/Views/BaseView.fs
Freymaurer: i noticed it updates everytime i switch the subpage, but we should switch to css:

`@media only screen and (max-width: XXXpx) {
 previousClass: {style}
}
`

kMutagene: Yes that is way cleaner. Can you open an issue for that so we have it in the backlog?

src/Server/Server.fs
Freymaurer: This works really well if you use examples. Otherwise the documentation lacks in comparison to OpenApi plus you cannot choose which Apis to document. The documentation for all api is automatic and you only need to fill in description and examples.

Issue #216, opened 2022-08-10
[BUG] Swate installer (for Windows) sometimes does not download manifests


omaus:
**Describe the bug**
When using Swate installer for Windows, this occurs in CMD window:
```
Created folder for Swate manifest @C:\Users\Kath9\AppData\Local\nfdi_manifests.
Download Swate manifest to folder..
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - Die Sperrfunktion konnte keine Sperrprüfung für das Zertifikat durchführen.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - Die Sperrfunktion konnte keine Sperrprüfung für das Zertifikat durchführen.
```
No manifests downloaded into manifests folder.

**To Reproduce**
(due to unknown reasons, it is not simply possible to reproduce this. It worked on several PCs without problems, but on some PCs it does not)

**Expected behavior**
curl.exe downloads the manifests without errors.

**OS and framework information (please complete the following information):**
 - OS: Win10

Freymaurer:
Can you create a PR fixing this issue on this branch https://github.com/nfdi4plants/Swate/tree/kevinf-patch-0.6 ?

PR #218, opened 2022-08-11
Previous commit from developer branch (curl.exe error)

omaus:
- cherry-pick 989f20fde03442bc06612783721971f4faf8c748 from omaus/Swate/developer
- see https://github.com/nfdi4plants/Swate/issues/216#issuecomment-1210730516 for background


Freymaurer (changes requested):
swate-win.zip contains assets folder
image
Should be only *.xml and install.cmd

Freymaurer (approved):
Nice, thank you!

PR #488, opened 2024-07-18
Feature, Excel, enable addition of building blocks, including terms and units

Etschbeijer:
Goal: 
* ~~Enable Addition of Terms and Units~~
* ~~Enable Addition of Building Blocks at selected Place~~
* Determine Distance beteween building blocks to insert new building blocks without overlapping

Problems:
* Determine the arc table index based on the excel table index
* This is needed to determine the type of building block that is selected
* Problem is, that we do not know whether it is a "Term" or "Unit", both are "isTerm = true"
* Another Problem is, we cannot check whether we have selected a "Main-" or "Propertycolumn" because the User can unhide those

### DIFF
--- a/src/Client/OfficeInterop/OfficeInterop.fs
+++ b/src/Client/OfficeInterop/OfficeInterop.fs
@@ -208,6 +208,9 @@ open System
 
 open Fable.Core.JsInterop
 
+[<Literal>]
+let AppendIndex = -1.
+
 /// <summary>This is not used in production and only here for development. Its content is always changing to test functions for new features.</summary>
 let exampleExcelFunction1 () =
     Excel.run(fun context ->
@@ -952,7 +955,7 @@ let addInputColumn (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColumn
 
         let rowCount = arcTable.RowCount + 1
 
-        let newColumn = ExcelHelper.addColumn -1. excelTable (newBB.Header.ToString()) rowCount ""
+        let newColumn = ExcelHelper.addColumn AppendIndex excelTable (newBB.Header.ToString()) rowCount ""
         let columnBody = newColumn.getDataBodyRange()
         // Fit column width to content
         columnBody.format.autofitColumns()
@@ -1025,7 +1028,7 @@ let addOutputColumn (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColum
 
         let rowCount = arcTable.RowCount + 1
 
-        let newColumn = ExcelHelper.addColumn -1. excelTable (newBB.Header.ToString()) rowCount ""
+        let newColumn = ExcelHelper.addColumn AppendIndex excelTable (newBB.Header.ToString()) rowCount ""
         let columnBody = newColumn.getDataBodyRange()
         // Fit column width to content
         columnBody.format.autofitColumns()
@@ -1074,10 +1077,10 @@ let addBuildingBlock (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColu
                 excelIndex + (float) targetIndex.Value + 1.
 
             //The +1 makes sure, we add the column after the column we have currently chosen
-            else -1.
+            else AppendIndex
         else
             //Add the new building block to the end of the table
-            -1.
+            AppendIndex
 
     let headers =
         headerRange.values[0]
--- a/src/Client/OfficeInterop/OfficeInterop.fs
+++ b/src/Client/OfficeInterop/OfficeInterop.fs
@@ -46,16 +46,16 @@ module OfficeInteropExtensions =
         /// <param name="rows"></param>
         static member viewRowsByColumns (rows:ResizeArray<ResizeArray<'a>>) =
             rows
-            |> Seq.collect (fun x -> Seq.indexed x)
+            |> Seq.collect (fun row -> Seq.indexed row)
             |> Seq.groupBy fst
             |> Seq.map (snd >> Seq.map snd >> Seq.toArray)
             |> Seq.toArray
 
-        static member addColumn (excelTable:Table) name rowCount value =
+        static member addColumn (index:float) (excelTable:Table) name rowCount value =
             let col = createMatrixForTables 1 rowCount value
             excelTable.columns.add(
-                index   = -1.,
-                values  = U4.Case1 (col),
+                index   = index,
+                values  = U4.Case1 col,
                 name    = name
             )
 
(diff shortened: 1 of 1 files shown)


src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -1038,6 +1038,80 @@ let addOutputColumn (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColum
...
+        let headers = arcTable.ToExcelValues().[0] |> Array.ofSeq |> Array.map (fun header -> header.ToString())
+
+        if excelIndex < (float) (headers.Length - 1) then 
+
+            //We want to start looking after the chosen column in order to ignore the current column because it could be a main column
+            let headers = headers.[(int) excelIndex + 1..]
+
+            let mainColumNames =
+                // Get all cases of the union
+                Microsoft.FSharp.Reflection.FSharpType.GetUnionCases(typeof<CompositeHeader>)
```
Freymaurer, line 1059: https://github.com/nfdi4plants/ARCtrl/blob/71a780f8df0290b0ede6d99858310521b439e2a2/src/Core/Table/CompositeHeader.fs#L159

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -1038,6 +1038,80 @@ let addOutputColumn (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColum
...
+                |> Array.map (fun case -> case.Name)
+
+            let indices =
+                mainColumNames
```
Freymaurer, line 1063: Not 100% sure whats happening here, but maybe do something like Array.choose with index

```fsharp
let headers : string [] = [|"Name"; "Age"; "City"|]
let mainColumnNames : string [] = [|"Name"; "Age"|] 

[
  for i in 0 .. headers.Length - 1 do
    let header = headers.[i]
    if mainColumnNames |> Array.exists (fun cName -> cName.StartsWith header) then
      i
]
// val it: int list = [0; 1]
```

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -1038,6 +1038,80 @@ let addOutputColumn (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColum
...
+        else
+            //Add the new building block to the end of the table
+            -1.
+
+    let headers =
+        headerRange.values[0]
+        |> List.ofSeq
+        |> List.map (fun header -> header.ToString())
+
+    let bb =
```
Freymaurer, line 1090: we do nothing with this? Use for loop and call sideffect excel functions inside. Therefore no new lists get instantiated for improved performance.

Freymaurer: You may also use `List.iteri`

Freymaurer (changes requested):

src/Client/OfficeInterop/OfficeInterop.fs
Freymaurer: I like the idea!

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed, removed bb
Was a rest from last testing sry

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -952,7 +952,7 @@ let addInputColumn (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColumn
 
         let rowCount = arcTable.RowCount + 1
 
-        let newColumn = ExcelHelper.addColumn excelTable (newBB.Header.ToString()) rowCount ""
+        let newColumn = ExcelHelper.addColumn -1. excelTable (newBB.Header.ToString()) rowCount ""
```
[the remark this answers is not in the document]
Etschbeijer: I don't understand what you mean, sry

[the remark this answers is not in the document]
Freymaurer:
```fsharp
[<Literal>]
let AppendIndex = -1
```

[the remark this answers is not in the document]
Etschbeijer: fixed

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -1056,38 +1059,35 @@ let addBuildingBlock (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColu
...
 
     let headers =
         headerRange.values[0]
         |> List.ofSeq
         |> List.map (fun header -> header.ToString())
 
-    let bb = 
+    let _ = 
         buildingBlockCells
         |> List.mapi(fun i bbCell ->
```
Freymaurer, line 1092: List.iteri, no binding. Otherwise we still instantiate a new list for performance loss

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -1096,12 +1096,12 @@ let addBuildingBlock (excelTable:Table) (arcTable:ArcTable) (newBB:CompositeColu
             let calIndex =
                 if targetIndex >= 0 then targetIndex + (float) i
                 else -1
+            log("targetIndex", targetIndex)
```
Freymaurer, line 1099: One of:

- Remove logs
- Make logs more declarative. Something like `log ("[AddBioldnigBlock] - Excel - <placeholder>")`

If you decide to make them more declarative write a excellog function which you can reuse for this purpose

Freymaurer (changes requested):

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed

PR #497, opened 2024-07-30
Feature excel delete building blocks

### LINKED ISSUE #495: [Excel]: Remove building block
### Contact Details

_No response_

### Is your feature request related to a problem?

In the quick access button list:

image

Make it so the selected building block with all related columns is removed.

### Describe the solution you would like!

.

### Additional context!

_No response_

### Host

Excel

### Host Details

.

Etschbeijer:
closes #495


src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -66,6 +66,27 @@ module OfficeInteropExtensions =
                 values  = U4.Case1 col
             )
 
+
+
+        static member adoptTableFormats (table:Table, context:RequestContext) =
```
Freymaurer, line 71: I had some function which did something similiar. This will be important for another one of the quickaccess buttons. Add a parameter which toggles hide or unhide.

if true -> hide all reference columns
if false -> unhide all columns

Freymaurer (changes requested):
The logic from `removeSelectedAnnotationBlock` seems overly complicated, are you sure this cannot be improved?

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed

Freymaurer (approved):

Issue #454, opened 2024-07-01
[Excel] Update Building block type + unit


Freymaurer:

Etschbeijer:
![grafik](https://github.com/user-attachments/assets/a4f21268-0b09-4e68-acc6-00c6dbaf1a5b)
Fronted adaption for excel needed, use swate as a basis

PR #502, opened 2024-08-05
Feature excel update term to unit

Etschbeijer:
Closses Issue #454

### DIFF
--- a/src/Client/SidebarComponents/Navbar.fs
+++ b/src/Client/SidebarComponents/Navbar.fs
@@ -36,17 +36,6 @@ let private shortCutIconList model dispatch =
                 SpreadsheetInterface.CreateAnnotationTable ctrl |> InterfaceMsg |> dispatch
             )
         )
-        QuickAccessButton.create(
-            "Update Term Building Block to Unit",
-            [
-                Html.i [prop.className "fa-solid fa-arrow-right-arrow-left pr-1"]
-                Html.i [prop.className "fa-solid fa-table-columns"]
-            ],
-            (fun e ->
-                e.preventDefault()
-                SpreadsheetInterface.UpdateUnitForCells |> InterfaceMsg |> dispatch
-            )
-        )
         match model.PersistentStorageState.Host with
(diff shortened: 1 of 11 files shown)


src/Client/SidebarComponents/Navbar.fs
```diff
@@ -36,6 +36,17 @@ let private shortCutIconList model dispatch =
                 SpreadsheetInterface.CreateAnnotationTable ctrl |> InterfaceMsg |> dispatch
             )
         )
+        QuickAccessButton.create(
```
Freymaurer, line 39: Converting between term and unit is a relatively rare usecase. I think it is better to keep the quicka ccess buttons for common use cases.

Etschbeijer: fixed

Etschbeijer:
The goal of this feature is the ability to convert building blocks.
You can convert terms into units and units into terms.
Additionally, you can also convert data into free text and free text into data.
You cannot convert input- and outpunt columns.

Etschbeijer:
The buttons to use the convert feature are located in the building blocks area of the swate tab.
grafik

Etschbeijer:
When you have selected a part of a building block in your table, you can klick on refresh. It will show you the type of building block you have chosen, e.g. Term. When it is convertible, you can click on "Convert XY to YX", converting the building block.
You do not have to click on refresh, you can directly click on "Convert XY to YX" and the button will be updated then.

Etschbeijer:
You can convert different building blocks with the implementation of this feature but the kind of building block you can convert to depends on the building block type.
You can check what type your current building block has by clicking on the "Refresh" button in the building block in the excel-swate-nav-bar.
grafik

Etschbeijer:
You can click he button below the refresh button to convert the currently selected building block into the building block type shown behind the button.
grafik

When you have selected another building block since you have selected the "Refresh" button, the buttons are not automatically updated. You can still convert your selected building block by pressen the "Convert" button, it will be updated then and show you what the building block used to be before you updated it.

PR #509, opened 2024-08-15
Feature excel add table validator

### LINKED ISSUE #452: [Excel] Table validation (ensure correct syntax, possible propagate ids)

Etschbeijer:
Closes #452


Etschbeijer:
Added a new button that checks whether the excel table is a valid arc/isa table or not.
When there is a problem, a message will be displayed for each occured problem, pointint at the column that is causing the problem for further investigation.
grafik
grafik

Freymaurer:
How would this handle a single column "Comment"?

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -194,11 +196,109 @@ module OfficeInteropExtensions =
...
                     )
                     return inMemoryTable
             }
 
+        /// <summary>
+        /// Checks whether the string seqs are part of a valid arc table or not and returns potential errors and index of them in the annotation table
+        /// </summary>
+        /// <param name="headers"></param>
+        /// <param name="rows"></param>
+        static member validateStringSeqs(headers:#seq<string>, rows:#seq<#seq<string>>) =
```
HLWeil, line 210: Give this member a better name. The name should include the info, that the "stringSeqs" represent a Table.

src/Client/OfficeInterop/OfficeInterop.fs
```diff
@@ -194,11 +196,109 @@ module OfficeInteropExtensions =
...
+            let columns = 
+                Seq.append [headers] rows 
+                |> Seq.transpose
+
+            let columnsArray = 
+                columns 
+                |> Seq.toArray
+                |> Array.map (Seq.toArray)
+
+            let groupedColumns = columnsArray |> ArcTable.groupColumnsByHeader
```
HLWeil, line 221: If you're going to validate the table column-wise anyways. Why not have a function which valdiates a single column and call that one sequentially here? This would increase readability I think.

HLWeil (changes requested):
Looks good!

Just a little readability changes to the code would be nice.

Also, would it make sense to distinguish between `Errors` and `Warnings`, instead of just warnings?

E.g. in your image example, what's the problem with column "Spalte3"? In general freetext columns are allowed. So this should not give the text "is invalid". But they're not good style, so a `Warning` might be okay.

On the other hand, when there is a column followed by a "Term Source REF" or "Unit" column, it MUST also itself conform to being a Parameter or something like it. So in this case an `Error` would be more suiting?

@Brilator @Etschbeijer ?

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: The problem is, I cannot really validate them column wise but based on the building blocks.
I have no Idea how you would validate a single column because a free text column can have any kind of information.
The only thing that could be validated whether the value in a term column is right or wront or am I overlooking something?

Etschbeijer:
> How would this handle a single column "Comment"?

@Freymaurer For some reason I cannot reply to your comment directly.
When it isn't breaking a building block there should not be a problem, unless the arc table says there is one.

Freymaurer:
> > How would this handle a single column "Comment"?
> 
> @Freymaurer For some reason I cannot reply to your comment directly.
> When it isn't breaking a building block there should not be a problem, unless the arc table says there is one.

Sorry it is difficult to review on phone. So you are using building block based arctrl parsing for this?

Etschbeijer:
> Sorry it is difficult to review on phone. So you are using building block based arctrl parsing for this?

@Freymaurer I thought it is the most important thing to be honest...

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: The problem with Spalte 3 is that it breaks a building block for Term and I was focussing on this. I was unsure how to validate otherwise and it seemed to be the most important part.

We could include more precise error warnings but I am using the ArcCtrl errors as a basis. For more precise warnings we would have to adapt or add functionality there when that is ok for you. I thought that might be a good thing for a future Issue. Now basic information that something is wrong and then mor precise information on the error in a future update.

Freymaurer:
> > Sorry it is difficult to review on phone. So you are using building block based arctrl parsing for this?
> 
> @Freymaurer I thought it is the most important thing to be honest...

Yeah dont worry! If you are using arctrl for this it is totally fine.

src/Client/OfficeInterop/OfficeInterop.fs
HLWeil: Oh okay I see,

Looking again into the code: I think, what's a bit confusing is that for every parsing error in the ARCTable you create two separate warnings.

src/Client/OfficeInterop/OfficeInterop.fs
HLWeil: Sry, single-column is not the correct term. I mean a column-group or single composite column.

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: Yes, I keep the original error from ArcCtrl and then point to the column that creates the error. It looks better than when I pack both of them into one warning (at least my impression).

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: The problem is, that the grouping method isn't working properly due to the false namings of the error and that makes a method/ function quite useless or not?

src/Client/OfficeInterop/OfficeInterop.fs
Etschbeijer: fixed the naming of the method

Etschbeijer:
In the building block tab the selected building block and those next to it can be validated, checking whether it can be part of an Arc / ISA table. Only the selected building block and those next to it can be check because it is not possible to group them correctly, when a header is false.
grafik

Freymaurer:
> In the building block tab the selected building block and those next to it can be validated, checking whether it can be part of an Arc / ISA table. Only the selected building block and those next to it can be check because it is not possible to group them correctly, when a header is false. grafik

This should not be a separate button, but be part of the "validate/auto fill" function in the quick access button bar

Etschbeijer:
grafik
By clicking on the Validate / Update Ontology Term button, currently, the table will be validated because the rest of the functionality of the button is not implemented yet.
You will be shown at least two messages, indicating which building block and in a more detailed message, which column is causing problems.
grafik

Etschbeijer:
> > In the building block tab the selected building block and those next to it can be validated, checking whether it can be part of an Arc / ISA table. Only the selected building block and those next to it can be check because it is not possible to group them correctly, when a header is false. grafik
> 
> This should not be a separate button, but be part of the "validate/auto fill" function in the quick access button bar

@Freymaurer Fixed

HLWeil (approved):

HLWeil (commented):
Still not sure about creating 2 separate warnings for a single problem in a column.

Etschbeijer:
@HLWeil Is it better than with two separate warnings?
grafik

HLWeil:
IMO yes, nice work!

PR #543, opened 2024-10-24
Add general form list drag and drop reorder #535

### LINKED ISSUE #535: [Feature Request] provide function to change the order of defined contacts
**Is your feature request related to a problem? Please describe.**
Usually researcher are interested in defining a specific author order. Currently one can not change the order of the contacts without deleting exisiting ones, e.g. adding a contact in the middle means deleting the last contacts and adding them again. 

**Describe the solution you'd like**
Maybe some simple  "+" and "-" buttons to move the contacts in the list up and down

Freymaurer:
closes #535

form-reorder

### DIFF
--- a/src/Client/SharedComponents/Metadata/Forms.fs
+++ b/src/Client/SharedComponents/Metadata/Forms.fs
@@ -286,38 +286,133 @@ module private Helper =
                 ]
             ]
         ]
-        
+
+open JsBindings
 
 type FormComponents =
 
     [<ReactComponent>]
-    static member InputSequence<'A>(inputs: ResizeArray<'A>, constructor: unit -> 'A, setter: ResizeArray<'A> -> unit, inputComponent: 'A * ('A -> unit) * (MouseEvent -> unit) -> ReactElement, ?label: string) =
-        Bulma.field.div [
-            if label.IsSome then Bulma.label label.Value
-            Bulma.field.div [
-                Html.orderedList [
-                    prop.className "grid grid-cols-1 gap-2"
+    static member InputSequenceElement(key: string, id: string, listComponent: ReactElement) =
+        let sortable = JsBindings.DndKit.useSortable({|id = id|})
+        let style = {|
+          transform = DndKit.CSS.Transform.toString(sortable.transform)
+          transition = sortable.transition
+        |}
+        Html.div [
+            prop.ref sortable.setNodeRef
+            prop.id id
+            for attr in Object.keys sortable.attributes do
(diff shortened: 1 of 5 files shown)


arendd:
Looks nice, functionality is great, only wondering about the icon with the "six dots" ? I'm not sure if this is obvious for the user that this if for drag&drop ? But only my personal impression.... THANK YOU

Freymaurer:
Good feedback! Do you have another icon in mind? We can easily use all free tier icons from https://fontawesome.com/

src/Client/SharedComponents/Metadata/Forms.fs
```diff
@@ -286,38 +286,126 @@ module private Helper =
...
 type FormComponents =
 
     [<ReactComponent>]
-    static member InputSequence<'A>(inputs: ResizeArray<'A>, constructor: unit -> 'A, setter: ResizeArray<'A> -> unit, inputComponent: 'A * ('A -> unit) * (MouseEvent -> unit) -> ReactElement, ?label: string) =
-        Bulma.field.div [
-            if label.IsSome then Bulma.label label.Value
-            Bulma.field.div [
-                Html.orderedList [
-                    prop.className "grid grid-cols-1 gap-2"
+    static member InputSequenceElement(key: string, id: string, listComponent: ReactElement) =
```
Etschbeijer, line 295: You have the parameter key that is never used in the method

Freymaurer: that is correct! it is not used in the method, but when transpiled to jsx this is similiar to

```tsx
<InputSequenceElement key={key} />
```

Which is the react required way keep react states on possible reorder events. [more](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key)

This looks strange but it is not necessary to use key inside the component, as simply passing the key parameter already sets key for react.

I actually only found out about this by looking into the Feliz.React source code. I am working on a PR to expand the Feliz docs accordingly

arendd:
What about one of these:
https://fontawesome.com/icons/arrows-up-down-left-right?f=classic&s=solid
https://fontawesome.com/icons/arrows-up-down?f=classic&s=solid

Freymaurer:
> What about one of these: https://fontawesome.com/icons/arrows-up-down-left-right?f=classic&s=solid https://fontawesome.com/icons/arrows-up-down?f=classic&s=solid

i updated to arrows-up-down 👍

Etschbeijer (approved):

Issue #544, opened 2024-10-25
[Feature Request]: add clickable link to ontology


Brilator:
### Is your feature request related to a problem?

Would help a lot to have a link to added terms in the Source Ontology.

### Describe the solution you would like!

I'd suggest to add this to the details widget

![Screenshot 2024-10-25 at 10 40 42](https://github.com/user-attachments/assets/67b60b70-0500-4e84-972a-10a3230e28ce)

### Host

Applies to All

### Host Details

all

Freymaurer:
> Would help a lot to have a link to added terms in the Source Ontology.

Sorry i don't quite understand 😞

Brilator:
How can I go from above swate annotation table to http://purl.obolibrary.org/obo/NCBITaxon_3702?

Freymaurer:
ahh okay, now i understand! Will do!

Freymaurer:
Done! Will be part of the next release!

Issue #539, opened 2024-10-23
Data Annotator: Absolute File Path in ARCitect


HLWeil:
### Reproduce
- Use Data Annotator in  ARCitect
- Pick file located in subfolder of dataset folder in assay 
  <img width="566" alt="image" src="https://github.com/user-attachments/assets/d8e7e782-b5d1-4171-aa30-d7cfb2d949d0">
- Pick cols and add 
-  Only file name is included 
  ![image](https://github.com/user-attachments/assets/a8389f4d-0af8-4400-97ae-c378078a5bad)

This is wrong by specification, as the file is not directly lcoated in the dataset folder. Even more problematic: Another file of the same name exists in another subfolder

### Favorite solution
I know this is tricky from a web-application perspective (i.e. in SWATE standalone), but would it be possible to retreive the necessary information in ARCitect?

HLWeil:
@frsommer

Freymaurer:
This should be implemented already. Can you verify that isa and target file are located in the same ARC?

Freymaurer:
Nevermind, just fixed it!

Issue #521, opened 2024-09-04
[Bug]: Change column header to "Characteristics" didn't work


SabrinaZander:
### Contact Details

sabrina.zander@hhu.de

### What happened?

I want to update a existing column from a "Parameter" to a "Characteristic". This didn't work. You can select "Characteristic" but the column header is not updated. If you want to change the column to any other term it works. 
Maybe there is a bug in the selection because "Characteristic" is the first term in the selection. 

### Host

Standalone

### Host Details

Standalone (Browser Google Chrome) 

### Swate version

I assume v.1.00 beta 07

Freymaurer:
Hello thanks for the feedback, i am currently still on vacation. Will look into this as soon as i get back.

For now you can fix this by editing the xlsx file by hand. I am sorry for the inconvinience!

SabrinaZander:
No problem, I only wanted to let you know.

Freymaurer:
I checked it and i think the issue is with `characteristic` being the default. The component checks applies updates on selection change. By clicking on `characteristic` it does not change as it is selected by default. If i click anything else and then `characteristic` it works. Can you verify this behavior on your side @SabrinaZander ? If yes, i will start working on a fix

SabrinaZander:
Yes, exactly

Freymaurer:
Done! will be live with the next release!

Issue #500, opened 2024-08-02
[Feature Request]: Display current file with file type


Freymaurer:
### Is your feature request related to a problem?

Difficult for user to easily understand the file they are currently working on.

### Describe the solution you would like!

Display in navbar? On questionmark icon?

### Host

Applies to All

### Host Details

.

Freymaurer:
![image](https://github.com/user-attachments/assets/f2440099-4881-4bc3-a1ea-c5ac64031f92)

Done, will be part of the next release

PR #549, opened 2024-10-28
Fixes to multiple smaller issues

Freymaurer:
- #500 
- #521
- #544 

Sadly #539 is still msising, as this will also require updates in ARCitect. Waiting for feedback on that one


src/Client/MainComponents/FooterTabs.fs
```diff
@@ -318,7 +318,7 @@ let MainMetadata(model: Model, dispatch: Messages.Msg -> unit) =
         prop.children [
             Html.a  [
                 Bulma.icon [Html.i [prop.className "fa-solid fa-circle-info"]]
-                Html.text "Metadata"
+                Html.text (model.SpreadsheetModel.FileType)
```
Etschbeijer, line 321: The Brackets are not needed

Etschbeijer (approved):

PR #566, opened 2024-11-13
Add manifest to swate public folder

Etschbeijer:


Freymaurer (changes requested):
Did you test this Manifest? As it still has the swate expert icons. A concept we more or less deprecated

Freymaurer:
> > Did you test this Manifest? As it still has the swate expert icons. A concept we more or less deprecated
> 
> @Freymaurer Yes, it worked nicely

Can you remove the references to swate_e_... icons and rename the swate addin from Swate.Core to Swate?

PR #586, opened 2024-12-05
Feature excel replace table name with template name

Etschbeijer:
Enables the takeover of the template that has been added to the table


Etschbeijer:
Swate Alpha:
grafik

PR #615, opened 2025-02-04
Search rework

Freymaurer:
- new term search :sparkles:
- Updated test setup
- Updated Component project
- Added storybook ui tests
- Added tib search support for new term search
- Added nuget package for Components project
- Added npm package for Components project
- Initial setup for publish workflows (current testing state for after merge into main)


src/Client/MainComponents/CellStyles.fs
```diff
@@ -1,107 +1,113 @@
...
-            Html.div [
-                prop.style [style.height(length.perc 100);]
-                prop.className "flex items-center justify-center px-2 py-1"
-                prop.disabled true
-                prop.children [
-                    Html.b (if rowIndex < 0 then "" else $"{rowIndex+1}")
+
+    static member CompositeCellDisplay (oa: OntologyAnnotation, displayValue: string) =
+        let hasValidOA = oa.TermAccessionShort <> ""
+        let v = displayValue
```
Etschbeijer, line 69: Why do you bind displayValue to v?

Etschbeijer (changes requested):

Etschbeijer:
grafik
Enable filling of multiple cells or adapt the text please

Etschbeijer:
grafik
Enable filling of multiple cells or adapt the text please

Etschbeijer:
grafik
Maybe relace the word "link" with "source"? It is a bit more descriptive
Move the icon with the cog to the top right, that is the same place we use it in the new Templates UI and so it looks a bit more in line.

src/Client/MainComponents/CellStyles.fs
Freymaurer: removed

Freymaurer:
> grafik Maybe relace the word "link" with "source"? It is a bit more descriptive Move the icon with the cog to the top right, that is the same place we use it in the new Templates UI and so it looks a bit more in line.

Renamed to `Source Link`, as `Source` already exists 50px above.

But I disagree with moving the cog to the top of the modal, as we do not allow to make adjustments to the current element, instead if functions as a display of used functions and should be lower priority. 

For example:

- Nobody scrolls to the bottom of all youtube videos to make adjustments to a search
- But when you are currently viewing a video, the video is in the focus and metadata can be found on the bottom

Freymaurer:
> grafik Enable filling of multiple cells or adapt the text please

Works as expected, allowing to fill multiple rows, but not multiple columns as the error message indicates:

image

Etschbeijer (approved):

Issue #640, opened 2025-02-25
[Bug]: When autosave is disabled local storage is still created


Etschbeijer:
### What happened?

I disabled autosave but a localstorage save was still created and slowed down the working process.

### What did you expect?

When the autosave is disabled, the localstorage shall not save a copy of the current file

### Host

Applies to All

### Host Details

Browser

### Swate version

v1.0.0-beta.25

Freymaurer:
Fixed will be live with the next release

PR #641, opened 2025-02-25
Bug: Disabled the saving of data in local storage when autosave is disabled

Etschbeijer:
Closses Issue #640


src/Client/Update/SpreadsheetUpdate.fs
```diff
@@ -39,6 +40,7 @@ module Spreadsheet =
             | _ ->
                 let newCmd =
                     if model.PersistentStorageState.Autosave then
+                        let snapshotJsonString = state.ToJsonString()
```
Freymaurer, line 43: `state.ToJsonString()` is one of the performance critical functions, and here it is called twice...

Freymaurer (changes requested):

Freymaurer (approved):

Issue #634, opened 2025-02-14
UI ISSUE COLLECTION


Freymaurer:
As we are currently supporting ARCitect, browser, vs code and excel we are encountering a lot of spacing ui issues with different window sizes.

this issue serves as a pinned collection for these issues.

Issue #629, opened 2025-02-11
[Bug]: Windows resizing


SabrinaZander:
### What happened?

Today during the workshop we had major problems for some participants with various windows in Swate (in ARCitect) that were not displayed correctly. This meant that you couldn't reach some buttons or could only see half of things. Here are some examples. It depends of course on the screen resolution.
It would be good to have the ability to resize the windows.

<img width="240" alt="Image" src="https://github.com/user-attachments/assets/6424d49c-2b2f-48a2-8540-b0ea9546e257" />

![Image](https://github.com/user-attachments/assets/37308f6c-703f-486e-b1ad-e5b83a059fbf)

Probably related to: 
https://github.com/nfdi4plants/Swate/issues/627 
https://github.com/nfdi4plants/Swate/issues/588

### Host

ARCitect

### Host Details

ARCitect version 0.0.52

### Swate version

v1.0.0-v1.0.0-beta.20

Freymaurer:
Can you give us the screen sizes? we will investigate this issue

Etschbeijer:
ARCitect:

![Image](https://github.com/user-attachments/assets/1e41d8be-18cf-45b6-9abb-3ba6a5965cc7)

![Image](https://github.com/user-attachments/assets/c2aba5ff-dd88-423e-8601-bc7b45a7dfae)

Freymaurer:
Fixed! Will be part of the next release!

PR #645, opened 2025-02-26
Widget bugs fixed 🐛 

Freymaurer:
Fixes 3 bugs with widgets:

- Bug 1: Widgets jump to 0 on x-axis as soon as the right screen window border touches them, like scared little kitties 🐈
- Bug 2: overflow for term search was set to scrollable
- Bug 3: Opening all widgets with the same params (Building Block, FilePicker, DataAnnotator), then clicking on them to get them in front made react not understand that they switched and reused their states (typical missing key issue). So i added a key to base widget and now they stay where they are supposed to stay.

### DIFF
--- a/src/Client/MainComponents/Widgets.fs
+++ b/src/Client/MainComponents/Widgets.fs
@@ -64,14 +64,13 @@ module private ResizeEventListener =
 
     open Fable.Core.JsInterop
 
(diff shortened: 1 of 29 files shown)


Freymaurer:
This should also fix some of the issues in #634 and #629

Etschbeijer (approved):

Issue #663, opened 2025-03-31
[BUG] Import JSON assay or study not working


frsommer:
**OS and framework information (please complete the following information):**
 - W10, ARCitect56, swate 1beta27

**Describe the bug**
- [ ] Want to import isa study or assay, does not import
- [ ] 
- [ ] export study or assay as json. new study or new assay, template, import json (study or assay, ISA), choose file, choose headers w values etc, choose sheets, press submit - nothing happens
- [ ] 
- [ ] screenshot and example assay file attached

![Image](https://github.com/user-attachments/assets/1065c46d-7246-4c37-81e4-def24ad28b23)

[Assay_SecondGet3TID_MS.json](https://github.com/user-attachments/files/19531190/Assay_SecondGet3TID_MS.json)

- [ ] should import selected values, worked before
- [ ] fix

Etschbeijer:
Hello @frsommer,

I am currently investigating the bug.
I found a small workaround, that enables you to import the data and work with it.

You do also have to import the metadata because by default it is not imported at the moment.

Do you also have trouble with the export or do I have a misunderstanding of your 2nd point?

Tell me please whether this workaround helped you or not for the moment.

frsommer:
Thanks for helping,

but the workaround does not work for me.

when importing an exported json-study including metadata as recomended:

![Image](https://github.com/user-attachments/assets/5434deb1-f50a-4218-9b39-b06b88a9cfae)

The sheet is imported, but lines are sorted in a different order:

![Image](https://github.com/user-attachments/assets/5276c841-73a4-4e95-a538-473d4579c7d7)

and when trying to save, an error occures since the Identifyer is not correct, i cannot find a place to change that:

![Image](https://github.com/user-attachments/assets/a91c2a69-a817-42cc-97dd-5e6034ee71c5)

Freymaurer:
Can you upload the study? I belive the file above is an assay? Then we will look into it 🙂

frsommer:
OK, it is getting even more confusing to me:
here are a couple of files

using arcitect 57, swate 1beta27, exported json:

assay:
[Assay_20250430_083710_APE1_TID_proteomics.json](https://github.com/user-attachments/files/19974779/Assay_20250430_083710_APE1_TID_proteomics.json)

And study:
[Study_20250430_083626_APE1_TID.json](https://github.com/user-attachments/files/19974782/Study_20250430_083626_APE1_TID.json)

**here the problem is** 
[...]

PR #671, opened 2025-05-07
Fixes import problem with assays and studys

Etschbeijer:
Fixes the import Issue for assays and co. reported in Issue #663.
Problem with the given file of Dr Sommer is not solved.


Freymaurer (approved):
lgtm

Issue #666, opened 2025-04-16
Context Menu


Freymaurer:

kMutagene:
still on this @Freymaurer or can this be moved to the backlog?

PR #702, opened 2025-06-04
Implement copy, cut, and paste functionality

Etschbeijer:
Part of Issue #666

### DIFF
--- a/src/Components/src/Table/AnnotationTableContextMenu.fs
+++ b/src/Components/src/Table/AnnotationTableContextMenu.fs
@@ -5,6 +5,19 @@ open Fable.Core.JsInterop
 open ARCtrl
 open Feliz
 
+[<AutoOpenAttribute>]
+module Helper =
+
+    type Clipboard =
+        abstract member writeText: string -> JS.Promise<unit>
+        abstract member readText: unit -> JS.Promise<string>
+
+    type Navigator =
+        abstract member clipboard: Clipboard
+
+    [<Emit("navigator")>]
+    let navigator: Navigator = jsNative
+
 /// AnnotationTableContextMenu Components
 type ATCMC =
     static member Icon (className: string) =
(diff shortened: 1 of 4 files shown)


src/Components/src/Table/AnnotationTableContextMenu.fs
```diff
@@ -15,19 +28,132 @@ type ATCMC =
...
     static member clear (tableIndex: CellCoordinate, cellIndex: (int * int), table: ArcTable, selectHandle: SelectHandle) : ArcTable =
         if selectHandle.contains tableIndex then
             table.ClearSelectedCells(selectHandle)
         else
             table.ClearCell(cellIndex)
         table.Copy()
 
+    static member deleteRow (tableIndex: CellCoordinate, cellIndex: (int * int), table: ArcTable, selectHandle: SelectHandle) : ArcTable =
+        if selectHandle.contains tableIndex then
+            let lastCellCoordinate =
```
Freymaurer, line 41: If the clicked on cell is contained in selected cells, ALL selected rows should be deleted not last

src/Components/src/Table/AnnotationTableContextMenu.fs
```diff
@@ -15,19 +28,132 @@ type ATCMC =
...
+            table.RemoveRow(lastCellCoordinate.y - 1)
+        else
+            table.RemoveRow(fst cellIndex)
+        table.Copy()
+
+    static member deleteColumn (tableIndex: CellCoordinate, cellIndex: (int * int), table: ArcTable, selectHandle: SelectHandle) : ArcTable =
+        if selectHandle.contains tableIndex then
+            let lastCellCoordinate =
+                selectHandle.getSelectedCells()
+                |> Array.ofSeq
```
Freymaurer, line 55: Same as rows. All should be deleted

src/Components/src/Table/AnnotationTableContextMenu.fs
```diff
@@ -15,19 +28,132 @@ type ATCMC =
...
+                        )
+                    )
+
+                CompositeCell.ToTableTxt(cells)
+            else
+                let cell = table.GetCellAt(cellIndex)
+                cell.ToTabStr()
+
+        navigator.clipboard.writeText result
+        |> Promise.start
```
Freymaurer, line 88: It is better style to call promise start wherewver this is used. As you can still add error handling on promises and manage timings

src/Components/src/Table/AnnotationTableContextMenu.fs
```diff
@@ -15,19 +28,132 @@ type ATCMC =
...
+
+    static member paste ((columnIndex, rowIndex): (int * int), table: ArcTable, selectHandle: SelectHandle, setTable) =
```
Freymaurer, line 90: This is incredibly complex. I am not yet sure how to correctly handle this. As we already got requests about only pasting values for unitized cells for example (keep unit, only change value).

It is very dofficult to read this part on phone. Can you give me a summary of the paste logic?

src/Components/src/Util/ARCtrl.Helper.fs
```diff
@@ -626,22 +626,35 @@ module Extensions =
                 | anyElse -> failwithf "Invalid content to parse to CompositeCell: %A" anyElse
 
         member this.ToTabStr() =
-            this.GetContentSwate() |> String.concat "\t"
+            if this.isFreeText then
```
Freymaurer, line 629: Is this handled correctly in excel? So the pasting in excel will paste a unitized term cell from swate into 4 excel columns?

src/Components/src/Util/ARCtrl.Helper.fs
```diff
@@ -626,22 +626,35 @@ module Extensions =
...
+                    |> String.concat "\t")
+            rows
+            |> String.concat (System.Environment.NewLine)
+
         static member fromTabTxt (tabTxt: string) (header: CompositeHeader) =
             let lines =
                 tabTxt.Split([| System.Environment.NewLine |], System.StringSplitOptions.None)
 
-            let cells = lines |> Array.map (fun line -> CompositeCell.fromTabStr line header)
+            let cells = lines |> Array.map (fun line -> CompositeCell.fromTabStr(line, header))
```
Freymaurer, line 657: Please write tests in Client.Tests for parsing cells to and from txt format!

This is a complex logic which we must ensure is handled correctly across different cases.

Freymaurer: Especially important to use roundabout tests, so check if swate cells -> tabtxt -> swate cells gives expected results.

If we must handle pasting with further user input we can check if reading tabtxt gives us the correct temp types

Freymaurer (changes requested):

src/Components/src/Util/ARCtrl.Helper.fs
Etschbeijer: grafik
A term looks like this, taking up 3 columns.

src/Components/src/Table/AnnotationTableContextMenu.fs
Etschbeijer: fixed

PR #733, opened 2025-06-16
Update tests

Etschbeijer:


tests/Client/Components/Fixture.fs
Freymaurer, line 1: maybe move this to `Components/Table/Fixture.fs`

Freymaurer (approved):
If you want a honest review, it's props best if you explaint to me via call

Freymaurer (changes requested):

Freymaurer (changes requested):

Freymaurer (approved):

Issue #715, opened 2025-06-11
[Bug]: The buger menu of swate is missing in ARCitect


Etschbeijer:
### What happened?

The buger menu in swate in ARCitect is missing.

### What did you expect?

The buger menu in swate in ARCitect is missing.

### Host

Applies to All

### Host Details

Windows 11

### Swate version

current

PR #771, opened 2025-07-17
Burger menu docs & contact open window with external browser

Etschbeijer:
Closes Issue #715


Freymaurer (commented):
Do these changes really change behavior of burger menu not displayed in ARCitect?

> The buger menu in swate in ARCitect is missing.

PR #812, opened 2025-08-15
Replace old basemodal with new one

Etschbeijer:


Freymaurer (changes requested):
i saw that you still passed `rmv` to most modals, but used `let open, setOpen = React.useState(true)` inside of the modals. In this case rmv, does not really do anything? or did i miss something?

Freymaurer (approved):

Issue #798, opened 2025-08-06
[Feature Request]: Escape when writing cell content should not save changes


omaus:
### Is your feature request related to a problem?

Like in MS Excel, I prefer Escape key aborting changes in cell content instead of saving them when writing.

### Describe the solution you would like!

Like in MS Excel, I prefer Escape key aborting changes in cell content instead of saving them when writing.

### Host

Applies to All

### Host Details

Current versions.

PR #819, opened 2025-08-26
When clicking escape keeps the old cell value

Etschbeijer:
Closes Issue #798


Freymaurer (changes requested):
This logic is too spread out. We want this behavior only inside table. So we should add something in base table implementation that tracks the preEdit value in the activeCellRender and propagates only under certain conditions. While omitting the changes under others

Issue #840, opened 2025-09-09
[Feature Request]: Improve table paste behavior


Freymaurer:
### Is your feature request related to a problem?

| Input [Data] | Output [Data] |
|--------| ------ |
| Input_1 | |
| Input_2 | |
| Input_3 | |

Try pasting Input_1 - Input_3 into the output column to replace input with "Edit" regex replace. Instead it will trigger: `Headers have been detected modal` forcing you to add a new column with `Input_1` as header or to cancel past.

### Describe the solution you would like!

Header detection should be more refined! `Input_1` should not be detected as header. Even then the header detected modal should offer "Paste into selected cells" option somehow. Instead of forcing user to past as header or to not paste

### Host

Applies to All

### Host Details

_._

PR #855, opened 2025-09-16
Fix check for header name

Etschbeijer:
Closes Issue #840

### DIFF
--- a/src/Components/src/AnnotationTable/AnnotationTableContextMenu.fs
+++ b/src/Components/src/AnnotationTable/AnnotationTableContextMenu.fs
@@ -83,7 +83,6 @@ type AnnotationTableContextMenuUtil =
         table.Copy()
 
     static member copy(cellIndex: CellCoordinate, table: ArcTable, selectHandle: SelectHandle) =
-
         let result =
             if selectHandle.getCount () > 1 then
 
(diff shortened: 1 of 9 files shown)


src/Components/src/AnnotationTable/AnnotationTableContextMenu.fs
```diff
@@ -230,12 +230,12 @@ type AnnotationTableContextMenuUtil =
 
             let areHeaders =
                 headers
-                |> Array.collect (fun header -> row |> Array.map (fun cell -> cell.StartsWith(header)))
+                |> Array.collect (fun header -> row |> Array.map (fun cell ->
+                    cell.StartsWith(header + " ") || cell = header))
```
Freymaurer, line 234: please use regex parsing from ARCtrl why try to reinvent the wheel

tests/Client/Components/Table/ContextMenu.tests.fs
Freymaurer, line 1: Still missing a new test of the issue. Remember to add the issue #-number to the test

Freymaurer (changes requested):

Freymaurer (changes requested):
I pulled the PR and found the following issues:

Assume a table of the format:

| Input [Source Name] | Parameter [temperature] | Output [Data] |
|---------------------|-------------------------|---------------|
| Output1             |                         |               |
| Output2             |                         |               |

## Wrong indexing

1. Select "Output1" & "Output2", use CtrlC or Copy via context menu. Verify correct copy in notepad.
2. Select single cell 0-0 (Output1). Returns "Cannot read properties of undefined (reading 'IsDataColumn')", this is due to indexing as trying to paste into Parameter [temperature] body cell returns "Invalid combination of header `Input [Source Name]` and cell `Output2`"

## Invalid combination

1. Follow steps from "Wrong indexing" and assume indexing issue is fixed.
2. Trying to paste "Output2" into "Input [Source Name] column throws with: "Invalid combination of header `Input [Source Name]` and cell `Output2`". As this is a freetext input and a freetext cell i would assume should work.

## Copy with headers does not work

1. Select full table inclusive headers and try to copy.
2. Throws with: "Uncaught Error: Unable to find cell for index: (0, -1) in table BraveDeer97"

I tried this to easily copy the table for this review and noticed that you cannot currently copy full tables. This should be supported to send other people tables in \t format, for bug reports and even for copy-pasting into excel.

src/Components/src/AnnotationTable/AnnotationTableContextMenu.fs
```diff
@@ -92,21 +91,42 @@ type AnnotationTableContextMenuUtil =
                     |> Array.ofSeq
                     |> Array.groupBy (fun item -> item.y)
 
-                let cells =
+                let headers, cells =
                     cellCoordinates
                     |> Array.map (fun (_, row) ->
                         row
                         |> Array.map (fun coordinate ->
```
Freymaurer, line 98: use Array.partition

Freymaurer: this will allow you to remove at least 20 lines of code

src/Components/src/Util/ARCtrl.Helper.fs
```diff
@@ -584,6 +584,38 @@ module Extensions =
             | CompositeHeader.Factor oa -> Some oa
             | _ -> None
 
+        member this.GetContentSwate() =
+            match this with
+            | CompositeHeader.Component oa -> [|this.ToString(); oa.NameText; defaultArg oa.TermSourceREF ""; defaultArg oa.TermAccessionNumber ""|]
```
Freymaurer, line 589: i am pretty sure this is wrong. 

for `CompositeHeader.Component oa` we would expect

| this.ToString() | Term Source REF (oa.TermAccessionShort) |Term Accession Number (oa.TermAccessionShort)|
|--------|--------|--------|
| Cell | Cell | Cell |

Please use existing ARCtrl logic something like this.

And: This does not work out if any of the cells for the header is a unitized cell. I think this function can only return correct results if you factor in the body cells too! 

Another example you have `Input io -> [|this.ToString()|]` which does only return one column, but when any body cell contains a Data cell you should return 3(4?) columns.

Etschbeijer: I am not using GetContentSwate and its parent method at the moment.
I wanted to talk with you about this, that is the reason why it is still in the code.

PR #856, opened 2025-09-16
Replaced new table with updating existing one

### LINKED ISSUE #853: [Bug]: Cannot add free text to cells in template
### Contact Details

_No response_

### What happened?

When creating/modifying a template, adding free text to a cell (not the column header) does not work when typing directly in the table. I can write something but the text is not kept. It only works when adding the term in the Add building block window.

### What did you expect?

_No response_

### Host

Standalone

### Host Details

Firefox

### Swate version

Version v1.0.0-rc.6

Etschbeijer:
Closes #853


Freymaurer (approved):
lgtm!

PR #858, opened 2025-09-17
Enable silent fail when api call termsearch fails

Etschbeijer:
Enable silent failing when a part of the termsearch fails


Freymaurer (approved):
lgtm!

PR #933, opened 2025-11-27
Keep FilePicker upload buttons visible for easier file management

caroott:
This PR ensures that the FilePicker upload controls are always rendered so users can overwrite existing selections without clearing the list

### DIFF
--- a/src/Client/Pages/FilePicker/FilePickerView.fs
+++ b/src/Client/Pages/FilePicker/FilePickerView.fs
@@ -258,22 +258,18 @@ type FilePicker =
 
     static member Main(model: Model, dispatch, containerQueryClass: string) =
         let hasFiles = not (List.isEmpty model.FilePickerState.FileNames)
+
         Html.div [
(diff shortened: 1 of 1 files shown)


src/Client/Pages/FilePicker/FilePickerView.fs
```diff
@@ -257,20 +257,23 @@ type FilePicker =
...
+                FilePicker.UploadButtons(model, dispatch, containerQueryClass)
+                if hasFiles then
                     FilePicker.FileSortElements model dispatch
                 Html.div [
                     prop.className "swt:overflow-y-auto swt:overflow-x-hidden swt:py-2"
                     prop.children [
-                        match model.FilePickerState.FileNames with
-                        | [] -> FilePicker.UploadButtons(model, dispatch, containerQueryClass)
-                        | _ -> FilePicker.FileViewTable model dispatch
+                        if hasFiles then
```
Freymaurer: Can propably remove this and just move this + the action buttons below to the `if hasFiles then` case in line 265

Freymaurer (approved):

PR #965, opened 2026-01-23
fix filetree and start adding swate components to preview

Etschbeijer:


src/Electron/src/Renderer/App.fs
Gitarackur: maybe if this can be moved into a new utility module (it seems that it can be reused across the project)

PR #971, opened 2026-02-04
Epic: ARCitect 2.0

Etschbeijer:


src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -283,9 +288,54 @@ let api: IArcVaultsApi = {
...
+                    vault.arc.Value.Workflows.Remove(workFlow)
+                    vault.arc.Value.Workflows.Add(updatedWorkflow)
+                    Ok()
+                | None -> Error(exn $"The ARC with the identifier {updatedWorkflow.Identifier} should exist")
+
+    updateARC =
+        fun _ ->
+            match ARC_VAULTS.TryGetVault(1) with
+            | None -> failwith $"The ARC for window id {1} should exist"
+            | Some vault -> vault.UpdateAsync()
```
caroott: `openFile`, `updateAssay`, `updateStudy`, `updateWorkflows`, and `updateARC `all call `ARC_VAULTS.TryGetVault(1)` instead of resolving the caller's window ID from an `IpcMainEvent`. The `openFile `API signature (`DataHolder `-> ...) doesn't even accept an event, so there's no way to derive the correct window. If a second window is opened via `openARCInNewWindow`, these operations will silently target the wrong vault.

src/Electron/src/Main/ArcVault.fs
```diff
@@ -271,6 +271,31 @@ module ArcVaultExtensions =
             | Some run -> Some run
             | None -> None
 
+        member this.GetContracts() =
+            match this.arc with
+            | Some arc -> arc.GetUpdateContracts()
+            | None -> failwith "No arc available"
+
+        member this.UpdateAsync() =
```
caroott: `UpdateAsync `calls both `arc.UpdateAsync` and `arc.WriteAsync` via `Promise.start`, returns unit, and the API type `updateARC`: `unit -> unit` propagates this. Caller cannot observe completion/errors.

src/Electron/src/Renderer/App.fs
```diff
@@ -108,6 +177,14 @@ let Main () =
         [| box previewData |]
     )
 
+    React.useEffect (
```
caroott: The `useEffect `depends on `box didSelectFile` and calls `updateARCVault` whenever `arcFileState.IsSome`. Since `openPreview` sets `didSelectFile = true` (lines 137, 142) and `CreateARCPreview` immediately resets it to false (line 33), each file click toggles the flag twice (false→true→false), triggering the effect twice per navigation, both times calling `updateARCVault`. The error path (line 142) also sets `didSelectFile = true`, so a failed file open also triggers a vault update attempt.

src/Electron/src/Renderer/App.fs
```diff
@@ -76,18 +75,88 @@ let ParseArcFileFromJson (fileType: ArcFileType) (json: string) : ArcFiles optio
...
+
+    Api.arcVaultApi.updateStudy(json)
+
+let updateWorkflow (workflow: ArcWorkflow) =
+
+    let json = workflow.ToJsonString()
+
+    Api.arcVaultApi.updateWorkflows(json)
+
+let updateARCVault (arcFiles: ArcFiles option) =
```
caroott: `updateARCVault` and `updateARC` both have | _ -> failwith "Not implemented yet" for Investigation, Run, DataMap, and Template. The Save button at MainElement.fs:  Line 94 is enabled whenever `arcFile.IsSome`, which includes those types. Clicking Save on an Investigation file will crash.

tests/Client/Electron/DataAnnotatorWidget.tests.fs
```diff
@@ -0,0 +1,128 @@
+module Electron.Tests.DataAnnotatorWidget
```
caroott: The test file has multiple incompatibilities with the implementation:

- Wrong namespace: imports `Renderer.components.DataAnnotatorWidget` but the namespace is `Renderer.components.Widgets`
- Wrong type path: uses `DataAnnotatorDataSource.AnnotationInput` (line 12) but `AnnotationInput` is a standalone type, not nested under `DataAnnotatorDataSource`
- Wrong casing: uses `selectorsFromTargets`, `tryGetTargetHeader`, `applyToTable`, `applyToDataMap`, `tryParseDataFile `(camelCase) while implementation uses PascalCase (`SelectorsFromTargets`, `TryGetTargetHeader`, `ApplyToTable`, `ApplyToDataMap`, `TryParseDataFile`)

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -283,9 +288,54 @@ let api: IArcVaultsApi = {
                     Swate.Components.console.log ("Unknown ISA file type, falling back to text preview")
 
                     try
-                        let content = fs.readFileSync (path, "utf8")
+                        let content = fs.readFileSync (newPath, "utf8")
                         return Ok(Text content)
                     with e ->
                         return Error(exn $"Could not read file {fileName}: {e.Message}")
         }
+    updateAssay =
```
caroott: `updateAssay`, `updateStudy`, `updateWorkflows` all do Remove(old) then Add(updated), which appends the updated item to the end instead of replacing in-place. This changes the serialized order of ISA entities.

src/Electron/src/Renderer/components/Widdgets/TemplateWidget.fs
```diff
@@ -0,0 +1,822 @@
...
+                                            ]
+                                        ]
+                                    ]
+                                ]
+                        ]
+                    ]
+            ]
+        ]
+
+    static member DropdownContent(isRowExtended: bool, setIsRowExtended) =
```
caroott: `DropdownContent`  renders a div with class `swt:collapse` containing only the checkbox and title. `SelectedColumnPreview` (line 402) renders with class `swt:collapse`-content but is a sibling of the collapse container (at TemplateWidget.fs Line 783-784), not a child. DaisyUI's collapse requires .collapse-content to be inside the .collapse element for the toggle to work.

src/Client/MainComponents/FooterTabs.fs
```diff
@@ -227,8 +227,9 @@ let MainPlus (model: Model, dispatch: Messages.Msg -> unit) =
             prop.onDragLeave <| dragleave_handler (state, setState)
             prop.onDragOver drag_preventdefault
             prop.onDrop <| drop_handler (order, state, setState, dispatch)
-            prop.onClick (fun e ->
```
caroott, line 230: e.ctrlKey was replaced with hardcoded false. The handler no longer captures the Ctrl key from the click event. Was this intentional?

src/Electron/src/Renderer/components/Widgets/WidgetHelper.fs
```diff
@@ -0,0 +1,43 @@
...
+
+type ActiveDataMapData = {
+    ArcFile: ArcFiles
+    DataMap: DataMap
+}
+
+type StatusElement =
+
+    [<ReactComponent>]
+    static member Create (status: StatusMessage) =
```
caroott: Identical implementations exist in BuildingBlockWidget, FilePickerWidget, and TemplateWidget.

src/Client/Pages/BuildingBlock/SearchComponent.fs
```diff
@@ -35,6 +35,30 @@ module SearchComponentHelper =
         |> InterfaceMsg
         |> dispatch
 
+    let addBuildingBlockARCitect (selectedColumnIndex: int option) (model: Model) dispatch =
```
caroott: functionally identical to `addBuildingBlock` (line 14). I see no usages of `addBuildingBlockARCitect` outside its definition.

src/Electron/src/Renderer/Renderer.fsproj
```diff
@@ -8,6 +8,12 @@
     <Compile Include="MetadataForm.fs" />
     <Compile Include="Api.fs" />
     <Compile Include="context\AppStateCtx.fs" />
+    <Compile Include="components\Widdgets\WidgetHelper.fs" />
```
caroott: Typo: Widdgets instead of Widgets

caroott (changes requested):
syncArcVault / SyncArcVault is copy-pasted in:

- App.fs:96 (updateARCVault — uses failwith for unsupported types)
- BuildingBlockWidget.fs:186 (| _ -> Ok () — silently ignores)
- DataAnnotatorWidget.fs:127 (| _ -> Ok () — silently ignores)
- FilePickerWidget.fs:98 (| _ -> Ok () — silently ignores)
- TemplateWidget.fs:126 (returns Error with messages for unsupported types)

--> Three different behaviors for the same operation.

caroott:
- `feature/StoreData` was rebased onto `origin/main` (which already included the landingPage merge), and save behavior from main was kept as the base.
- Save serialization/parsing/preview mapping was deduplicated into shared code (`ArcFileSaveMapping`) and main IPC was updated to use it while keeping main-specific ARC mutation checks.
- Renderer-side redundant widget save paths were removed and unified through a dedicated `Renderer/ArcFilePersistence.fs` module.
- Additional cleanup refactored `openFile`/landing preview construction in main to reuse shared mapping, and fixed DataMap open to return actual ARC DataMap content.
- Renderer `Api.fs` got wrappers for ARC window operations, and `App.fs` switched to those wrappers.

src/Electron/src/Renderer/components/Widgets/BuildingBlockWidget.fs
Freymaurer: Any "components" should be placed in `src/components` to enable isolated tests and releases as npm package. This is true for all components/widgets

PR #987, opened 2026-02-25
Added a Search componenet for the Notes including story notebook

Rookabu:
This is a simple search component which allows the user to search in their notes and open them. Tested on 10 test notes. 

image

Clicking into the search field before typing suggests all notes.

image

 If the user types in characters, the search will suggest notes where that string appears, independent from lower and upper case.

image

The "open note" button should later trigger the actual note functionality, allowing to view and edit them (?)

image

### DIFF
--- a/src/Components/playground/App.tsx
+++ b/src/Components/playground/App.tsx
@@ -18,6 +18,7 @@ import {Entry as DataMapTable} from '../src/DataMapTable/DataMapTable.fs.ts';
 import {Entry as Layout} from '../src/Layout/Layout.fs.js'
 import {FileExplorerExample_Example} from '../src/FileExplorer/FileExplorer.fs.ts'
 import {Entry as WidgetController} from '../src/Widgets/Widgets.fs.ts';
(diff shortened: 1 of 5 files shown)


Freymaurer:
1. What does the Toggle Search button do?
2. You have some overlapping border and spacong issues. Also the single notes do all have borderradius, instead of a grouped element feel.

Rookabu:
The toggle button is simply for opening/closing the whole search since it will be  also triggered via a button later in swate. You're right with the missing grouped element feel, I`ll fix this. @Freymaurer

Freymaurer:
> The toggle button is simply for opening/closing the whole search since it will be  also triggered via a button later in swate. You're right with the missing grouped element feel, I`ll fix this. @Freymaurer 

If you want you can remove the open/close part from the component. I am not sure how and where we will add this component exactly. So you can focus and the browser itself 🙂

Rookabu:
@Freymaurer

Etschbeijer (approved):
lgtm

Issue #980, opened 2026-02-23
[Feature Request]: Enable usage of Git LFS in file tree


Etschbeijer:
### Is your feature request related to a problem?

Currently Git LFS is not implemented

### Describe the solution you would like!

* Enable marking of LFS files with the context menu in File Explorer using Git-LFS

### Host

Applies to All

PR #988, opened 2026-02-25
 Git LFS

Gitarackur:
This PR adds Git LFS integration for Electron File Explorer context menus and wires renderer actions to main-process Git commands.

Addresses issue #980


src/Components/src/FileExplorer/FileTreeDataStructures.fs
```diff
@@ -421,3 +422,15 @@ module FileExplorerLogic =
                     model.Items
 
             { model with Items = newItems }
+
+        | SetLFSMarked(itemId, isMarked) ->
+            let newItems =
+                FileTree.updateItem
+                    itemId
+                    (fun item -> {
```
caroott: Looks to me like IsLFS is set immediately, but the async call can still fail afterwards. It then returns only an error message and doesn't change this back (FileExplorer.fs Line~100

src/Electron/src/Main/IPC/GitLfs.fs
```diff
@@ -0,0 +1,254 @@
...
+        | Fetch, Some file -> Ok [ "lfs"; "fetch"; "--include"; file ]
+        | Install, _ -> Ok [ "lfs"; "install" ]
+        | Track, Some file -> Ok [ "lfs"; "track"; file ]
+        | Untrack, Some file -> Ok [ "lfs"; "untrack"; file ]
+        | Status, Some file -> Ok [ "lfs"; "ls-files"; "--name-only"; "--"; file ]
+        | Track, None
+        | Untrack, None
+        | Status, None -> Error "FilePath is required for this Git LFS command"
+
+    let validateRepoPath repoPath =
```
caroott, line 43: This currently only checks, if it's a git repo. Should be fine for all ARCitect use cases, but if somehow another git repo path than the opened arc manages to "sneak in", this would return true

src/Electron/src/Main/IPC/GitLfs.fs
```diff
@@ -0,0 +1,254 @@
...
+    let mutable isRunning = false
+
+    let toArgs (request: GitLfsRequest) =
+        match request.Command, request.FilePath with
+        | Pull, None -> Ok [ "lfs"; "pull" ]
+        | Pull, Some file -> Ok [ "lfs"; "pull"; "--include"; file ]
+        | Fetch, None -> Ok [ "lfs"; "fetch" ]
+        | Fetch, Some file -> Ok [ "lfs"; "fetch"; "--include"; file ]
+        | Install, _ -> Ok [ "lfs"; "install" ]
+        | Track, Some file -> Ok [ "lfs"; "track"; file ]
```
caroott: git lfs track/untrack path arg is missing `--` separator. Track/Untrack pass raw filename as argument. Filenames starting with `-` can be parsed as options by git-lfs

src/Components/src/FileExplorer/FileExplorer.fs
```diff
@@ -56,27 +56,36 @@ type FileExplorer =
...
                     let windowObj: obj = Browser.Dom.window
-                    do! windowObj?navigator?clipboard?writeText(path)
+                    do! windowObj?navigator?clipboard?writeText (path)
                 with ex ->
                     Browser.Dom.console.warn ($"Could not copy file path: {path}", ex)
             }
             |> Promise.start
 
         let defaultContextMenuItems (item: FileItem) : ContextMenuItem list =
+            let isLfsMarked = item.IsLFS = Some true
```
caroott: This now hardcodes Git LFS menu/actions/text, which conflicts with the “application-agnostic reusable UI” guideline for components in Structure.md

Gitarackur: LFS context menu has now been isolated from the Swate components and moved to the renderer

src/Electron/src/Main/IPC/GitLfs.fs
Gitarackur: added

src/Electron/src/Renderer/components/FileExplorer.fs
```diff
@@ -78,6 +93,35 @@ let createFileTree (parent: FileItemDTO option) selectedTreeItemPath setSelected
 
     let fileItem = loop parent
 
+    let toggleLfsMark (item: FileItem) (markAsLfs: bool) =
```
Etschbeijer: Can you move this to a fileexplorer - git-LFS helper type in Swate-Components please?

Etschbeijer (changes requested):

src/Components/src/FileExplorer/FileTreeDataStructures.fs
Gitarackur: done

src/Electron/src/Renderer/components/FileExplorer.fs
Gitarackur: as a utility function, right? @Etschbeijer

src/Electron/src/Renderer/components/FileExplorer.fs
```diff
@@ -71,13 +87,72 @@ let createFileTree (parent: FileItemDTO option) selectedTreeItemPath setSelected
...
+
+    let contextMenuItems (item: FileItem) =
+        if item.IsDirectory then
+            []
+        else
+            let isMarked = item.IsLFS = Some true
+
+            [
+                {
+                    Label = if isMarked then "Unmark Git LFS" else "Mark as Git LFS"
```
caroott: isLfs source does not match Track/Untrack semantics, so UI status can still be wrong.
FileTreeCreator now marks files using git lfs ls-files --name-only (FileTreeCreator.fs Line 12, 54).
That command reports LFS pointers in index/history, not purely .gitattributes track rules. So after track/untrack, the context label in renderer can still be stale/inverted depending on staging state.

src/Electron/src/Main/FileTreeCreator.fs
```diff
@@ -7,6 +7,25 @@ open Swate.Electron.Shared.IPCTypes
 
 let fs: obj = importAll "fs"
 let pathMod: obj = importAll "path"
+let childProcessDynamic: obj = importAll "node:child_process"
+
+let private tryGetLfsTrackedPaths (repoPath: string) =
```
caroott: someting like this:
```
let isLfsTracked (repoRoot: string) (absolutePath: string) =
    try
        let relativePath =
            pathMod?relative(repoRoot, absolutePath)
            |> unbox<string>
            |> fun p -> p.Replace("\\", "/")
        if System.String.IsNullOrWhiteSpace relativePath || relativePath = "." then
            false
        else
            let output: string =
                childProcessDynamic?execFileSync(
                    "git",
                    [| "check-attr"; "filter"; "--"; relativePath |],
                    createObj [
                        "cwd" ==> repoRoot
                        "encoding" ==> "utf8"
                        "stdio" ==> "pipe"
                        "shell" ==> false
                    ]
                )
                |> unbox<string>
            output.Contains(": filter: lfs")
    with _ ->
        false
```

src/Electron/src/Main/FileTreeCreator.fs
```diff
@@ -15,31 +34,37 @@ let getFileEntry (path: string) =
...
                 pathMod?basename(path),
                 path,
-                stats?isDirectory()
+                stats?isDirectory(),
+                None
             )
     }
 
 ///Finds all files and subfolders of the given filepath
-let rec getFileEntries (path: string) =
```
caroott, line 23: and this:
```
/// Finds all files and subfolders of the given filepath
let getFileEntries (path: string) =
    let repoRoot = path.Replace("\\", "/")

    let rec loop (currentPath: string) =
        let stat = fs?statSync(currentPath)
        let isDir = stat?isDirectory()
        let name = pathMod?basename(currentPath)
        let normalizedPath = currentPath.Replace("\\", "/")

        let isLfs =
            if isDir then
                None
            else
                Some(isLfsTracked repoRoot normalizedPath)

        let currentEntry =
            IPCTypes.FileEntry.create(name, normalizedPath, isDir, isLfs)

        if isDir then
            let entries: string[] = fs?readdirSync(normalizedPath)

            let children =
                entries
                |> Array.collect (fun entry ->
                    let fullPath = pathMod?join(normalizedPath, entry)
                    loop fullPath
                )

            Array.append [| currentEntry |] children
        else
            [| currentEntry |]

    loop repoRoot

```

src/Electron/src/Main/FileTreeCreator.fs
Gitarackur: updated

src/Electron/src/Renderer/components/FileExplorer.fs
caroott: - `git lfs track` / `git lfs untrack` changes tracking rules in .gitattributes.
- `git lfs ls-files` shows files that are currently LFS pointer objects in index/commits (from any source)

`isLfs` is populated from `git lfs ls-files --name-only` . Then the renderer uses that for menu text. That means UI text is effectively saying “is pointer right now”, while wording implies “is tracked by rule”.

Mismatch Example:

Existing pointer file, then `git lfs untrack file.txt`:
- rule is removed, but ls-files can still list pointer until content/index is rewritten
- UI can show “marked” even though rule is now untracked

`git check-attr filter -- <file>` reports what `.gitattributes` currently applies to that path

Etschbeijer (approved):
lgtm

Issue #999, opened 2026-03-02
[Feature Request]: Enable authentification logic for log in of ARCitect v2


Etschbeijer:
### Is your feature request related to a problem?

Currently ARCitect v2 works only locally.

### Describe the solution you would like!

Authentification logic:
* When you are not logged in use a sign up button, otherwise a sign in button. Both point to the adjacent dataplant pages, working as popover utilizing a classic "user-Portrait"
* The login utilizes a forward of the token on an extra localhost. Locale storing of the token shall be done utilizing best practices (still looking up).
* Testing of the login with the help of a private repo (manually for now - no end to end required here).
* Enable selection of the databub with a multiselect. Dataplant shall be its default choice.

### Host

Applies to All

PR #1000, opened 2026-03-02
Epic/store data/auth #999

Freymaurer:
Components were designed following:
- Component isolation
- flexibel callbacks for sign in and logout
- user state is stored outside of main components
- minimal component tests were implemented
- contribution nr. 1000! 🚀

image

image

---

I further added some prototypes for gitlab browser and git components:

image

### DIFF
--- a/src/Components/src/Authentication/Authentication.fs
+++ b/src/Components/src/Authentication/Authentication.fs
@@ -1,0 +1,493 @@
+namespace Swate.Components
+
+open Fable.Core
+open Fable.Core.JsInterop
+open Feliz
+
+open AuthenticationTypes
+
+
+type private DataHubInformation = {
+    Name: string
+    Url: string
+    Description: string option
+} with
+
+    member this.GetDescription() =
+        match this.Description with
+        | Some description -> description
+        | None -> this.Url
+
+
+type private DataHubCollection = {
+    Default: DataHubInformation
+    Supported: DataHubInformation list
+}
+
+module GitLabAPI =
+
+    open Fetch
+
+    [<RequireQualifiedAccess>]
+    type SignInError =
+        | NetworkError of exn
+        | Unauthorized
+        | Forbidden
+        | NotFound
+        | HttpError of int
+        | DecodeError of exn
+
+    let getUserAPIRequest (signInInfo: SignInInformation) : JS.Promise<Result<UserInformation, SignInError>> = promise {
+        let baseUrl = signInInfo.GitLabBaseUrl.TrimEnd('/')
+        let pat = signInInfo.PersonalAccessToken
+        let url = $"{baseUrl}/api/v4/user"
+
+        let requestOptions = [
+            RequestProperties.Method HttpMethod.GET
+            requestHeaders [ HttpRequestHeaders.Custom("PRIVATE-TOKEN", pat) ]
+        ]
+
+        try
+            console.log ($"Making API request to {url} with PAT")
+            let! response = fetchUnsafe url requestOptions
+
+            // ---- HTTP STATUS HANDLING ----
+            if not response.Ok then
+                return
+                    match response.Status with
+                    | 401 -> Result.Error SignInError.Unauthorized
+                    | 403 -> Result.Error SignInError.Forbidden
+                    | 404 -> Result.Error SignInError.NotFound
+                    | code -> Result.Error(SignInError.HttpError code)
+            else
+
+                // ---- JSON PARSE ----
+                try
+                    let! gitLabUserInfo = response.json<AuthenticationTypes.GitLabUser> ()
+
+                    let userInfo =
+                        UserInformation.FromGitLabUser gitLabUserInfo pat signInInfo.GitLabBaseUrl
+
+                    return Ok userInfo
+
+                with ex ->
+                    return Error(SignInError.DecodeError ex)
+
+        // ---- NETWORK ERROR ----
+        with ex ->
+            return Error(SignInError.NetworkError ex)
+    }
+
+module private AuthenticationHelper =
+
+    let prefillGitLabPATScopes (gitlabBaseUrl: string) =
+        let gitlabBaseUrl = gitlabBaseUrl.TrimEnd('/')
+
+        let scopes = [
+            "api"
+            "read_user"
+            "read_repository"
+            "read_api"
+            "write_repository"
+        ]
+
+        let scopeParam = scopes |> String.concat ","
+
+        let description =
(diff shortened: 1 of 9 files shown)


src/Components/src/Authentication/Authentication.fs
```diff
@@ -0,0 +1,493 @@
...
+        with ex ->
+            return Error(SignInError.NetworkError ex)
+    }
+
+module private AuthenticationHelper =
+
+    let prefillGitLabPATScopes (gitlabBaseUrl: string) =
+        let gitlabBaseUrl = gitlabBaseUrl.TrimEnd('/')
+
+        let scopes = [
```
caroott, line 86: Those scopes are are maybe a bit overpriviledged. GitLab API access has permissions beyond repo read/write afaik

src/Components/src/Authentication/Authentication.fs
```diff
@@ -0,0 +1,493 @@
...
+            "read_repository"
+            "read_api"
+            "write_repository"
+        ]
+
+        let scopeParam = scopes |> String.concat ","
+
+        let description =
+            "Swate Electron App. Gives access to your repositories and allows Swate to read your user information. This is used to authenticate you and access your ARCs. You can revoke this token at any time without affecting any other tokens or your account."
+                .Replace(" ", "%20")
```
caroott, line 97: &/#/non-ASCII may break this replacement

src/Components/src/Authentication/Authentication.fs
```diff
@@ -0,0 +1,493 @@
...
+                            prop.className "swt:loading swt:loading-spinner swt:loading-sm"
+                        ]
+                    | _, Some userInformation -> Authentication.UserAvatarIcon(userInformation)
+                    | _, None -> Authentication.AvatarPlaceholder()
+                ),
+                [| box userInformation; box isLoading |]
+            )
+
+        let content =
+            React.useMemo (
```
caroott, line 410: Memoized tree closes over onLogout/onSignIn. Any other actions (e.g. crash?) could maybe lead to stale handlers

src/Components/src/Authentication/Types.fs
```diff
@@ -0,0 +1,76 @@
...
+type GitLabIdentity = {
+    provider: string
+    extern_uid: string
+    saml_provider_id: int option
+}
+
+type GitLabScimIdentity = obj
+// Empty array in sample; replace with proper type if structure becomes known
+
+type GitLabUser = {
```
caroott, line 18: very broad and most not optional

src/Components/src/Authentication/Authentication.fs
```diff
@@ -0,0 +1,493 @@
...
+            "%s/-/user_settings/personal_access_tokens?name=swate-electron&description=%s&scopes=%s"
+            gitlabBaseUrl
+            description
+            scopeParam
+
+    [<Literal>]
+    let Default_DataHub_Url = "https://git.nfdi4plants.org/"
+
+    let Default_DataHub = {
+        Name = "PLANTdataHUB (offical)"
```
caroott: what is this offical you are talking about?

src/Components/src/Authentication/Authentication.fs
```diff
@@ -0,0 +1,493 @@
...
+                            prop.className "swt:text-base-content swt:text-lg swt:font-semibold"
+                            prop.text dataHub.Name
+                        ]
+                        Html.p [
+                            prop.className "swt:text-xs swt:text-base-content/70"
+                            prop.text (dataHub.GetDescription())
+                        ]
+                        Html.a [
+                            prop.href dataHub.Url
+                            prop.target "_blank"
```
caroott, line 237: This can be abused if dataplant gets hacked 😨
`prop.target="_blank"` page can use window.opener to navigate to malicious sites.
Adding `prop.rel "noopener noreferrer"` prevents that

src/Components/src/Authentication/Authentication.fs
Freymaurer: Which scopes would you suggest?

src/Components/src/Authentication/Authentication.fs
Freymaurer: Is this a warning for future changes? Or does this affect the current description? This somewhat sound sassy but i am really not sure if i might not see something 😅

src/Components/src/Authentication/Authentication.fs
Freymaurer: I'll add error to dependencies

src/Components/src/Authentication/Types.fs
Freymaurer: What do you mean with this comment? What changes would you like to see?

caroott: maybe make them options or remove them if they are not used

src/Components/src/Authentication/Authentication.fs
caroott: future changes

src/Components/src/Authentication/Authentication.fs
caroott: For simple-git we currently only need `read_repository` for clone/fetch/pull and `write_repository` for push over HTTPS.
`read_user`, `read_api`, and `api` are currently not required. Maybe you need `read_user` for your functions, but probably not the api scopes

Freymaurer: I assume `api` scope would be something along the lines of "Give me a list of all my personal repositories/arcs"? Which is a feature we will need.

src/Components/src/Authentication/Authentication.fs
Freymaurer: In this case i would ignore it, as this is a static text, and should be adjusted when relevant. If you agree?

src/Components/src/Authentication/Types.fs
Freymaurer: I created this type, based on my personal user info i got. From an account that had nearly no information given, as i do not really use it. So i think this should be correct. But i can add a link to the gitlab api ref

src/Components/src/Authentication/Authentication.fs
Freymaurer: i removed api, and we will see if we require it again :)

src/Components/src/Authentication/Authentication.fs
Etschbeijer: lgtm

Etschbeijer (approved):
lgtm

PR #1001, opened 2026-03-02
Added filter for search

Rookabu:
Search suggestion in notes can now be filtered if the user wants to search in tags, title, content or all. 

grafik

dropdown open:

grafik

grafik

grafik

Text if no notes are found: 

grafik


Etschbeijer (changes requested):
Problem:
Search in dropdown changes its size, depending on the chosen condition.

Etschbeijer (approved):
lgtm

PR #1005, opened 2026-03-04
make filetree async and filter .git from filetree

caroott:
Make ARC file tree generation faster by switching to async traversal, batching Git LFS checks, and excluding `.git` from scans.

## Changes
- Refactor `getFileEntries` to use `fs.promises` + iterative traversal (no sync IO / deep recursion).
- Batch LFS detection via `git check-attr -z filter --stdin` instead of per-file checks.
- Filter `.git` from both the file tree and file watcher ignore rules; skip temp Excel lock files (`.~$*.xlsx`).
- Update vault/IPC call sites to `await` file entry collection before building the tree.


Etschbeijer (approved):
lgtm

PR #1009, opened 2026-03-11
Add Auth workflow to electron app

Freymaurer:
- multi-datahub login/logout workflow
- using safeStorage for personal access token storage
- new ipc controller for auth workflows

#999


Etschbeijer (approved):
lgtm

PR #1017, opened 2026-03-18
Use deterministic SHA-256 account IDs to prevent auth account collisions

caroott:
`SecureAuthStore.generateAccountId` generated account IDs by replacing non-alphanumeric characters with underscores. That allowed distinct emails to collapse to the same `accountId`, for example:

- `a.b@x`
- `a_b@x`

Both could map to the same ID

## Fix

Replace the sanitize-based account ID generation with a deterministic SHA-256 hash based on the normalized account identity:

- normalized DataHub host
- trimmed, lowercased email

The generated `accountId` is now a stable 64-character hex hash, which avoids collisions caused by character replacement.

## Additional Changes

- Added tests covering:
  - the previous collision case
  - normalization behavior
  - filesystem-safe hash output


src/Electron/src/Main/Auth/SecureAuthStore.fs
```diff
@@ -33,6 +33,28 @@ let private bufferToBase64 (buf: Node.Buffer.Buffer) : string = jsNative
...
+
+[<Import("createHash", "crypto")>]
+let private createHash (algorithm: string) : CryptoHash = jsNative
+#endif
+
+let private sha256Hex (input: string) : string =
+#if FABLE_COMPILER
+    let hash = createHash "sha256"
+    hash.update(input, "utf8").digest("hex")
+#else
```
Freymaurer: this project cannot run in dotnet, as the full framework only runs on node. So no need for compiler directives

tests/Server/SecureAuthStore.Tests.fs
```diff
@@ -0,0 +1,36 @@
+module SecureAuthStoreTests
```
Freymaurer: These tests should be in a fable-transpiled only test project. To keep things cleanly separated: Please create a new test project in tests/Electron.Core

Use any of the following testing frameworks:
- Pyxpecto
- Vitest (fable, f# variant)
- Fable.Mocha

If you have problems setting this up feel free to call me :)

Freymaurer (changes requested):

Etschbeijer (approved):
lgtm

PR #1018, opened 2026-03-18
Epic/store data/gitlab browser

Freymaurer:
image


src/Components/src/DataHub/DataHubBrowser.fs
```diff
@@ -0,0 +1,1328 @@
...
+                    state with
+                        SelectedGroupId = selectedGroupId
+                        Page = 1
+                }
+        | SetPage page ->
+            if page = state.Page then
+                state, Cmd.none
+            else
+                reloadWith { state with Page = page }
+        | LoadReposRequest request ->
```
caroott, line 139: `LoadReposResponse` is applied blindly, with no request id / version check. If a slower older GitLab request finishes after a newer tab/search/sort request, the stale response can overwrite the current state.

src/Components/src/DataHub/DataHubBrowser.fs
```diff
@@ -0,0 +1,1328 @@
...
+                match request.SelectedGroupId with
+                | Some gid -> Some gid
+                | None when groups.Length > 0 -> Some groups[0].id
+                | None -> None
+
+            let! result =
+                match request.Target with
+                | ExploreTab.All -> loaders.LoadAllRepos repoQuery
+                | ExploreTab.YourRepos ->
+                    if request.IsAuthenticated then
```
caroott, line 690: The organisations flow already falls back to `groups[0]` inside `loadRepos` and uses it for the first `LoadOrganisationRepos` call. When that response comes back, `LoadReposResponse` notices `state.SelectedGroupId = None` and triggers a second reload for the same group. First entry into the tab therefore does `groups + repos + repos`, and the blank "Select organisation" option can never remain selected.

src/Components/src/DataHub/DataHubBrowser.fs
```diff
@@ -0,0 +1,1328 @@
...
+                SearchTerm = request.SearchTerm
+                Page = request.Page
+                PerPage = request.PerPage
+                OrderBy = orderBy
+                Sort = sort
+                Visibility = visibility
+            }
+
+            let! groupsResult =
+                if request.Target = ExploreTab.YourOrganisations && request.IsAuthenticated then
```
caroott, line 670: The organisation picker is hard-capped to `Page = 1; PerPage = 100`, and the pagination metadata is thrown away. GitLab paginates `/groups`, so users with more than 100 accessible groups will never be able to browse the remaining groups from this UI.

src/Components/src/DataHub/Types.fs
```diff
@@ -0,0 +1,137 @@
...
+    | LastUpdated
+    | DateCreated
+    | Name
+    | Stars
+
+module ExploreSortField =
+
+    let toProjectSortField =
+        function
+        | ExploreSortField.LastUpdated -> ProjectSortField.UpdatedAt
```
caroott, line 24: `ExploreSortField.LastUpdated` maps to `ProjectSortField.UpdatedAt`, but the row UI renders `last_activity_at` as the visible "Updated ..." timestamp, and the mock entry sorts by `last_activity_at`. On real GitLab data the production order can therefore disagree with the timestamp shown to the user, while the stories/tests still pass because the mock path uses different semantics.

src/Components/src/DataHub/DataHubBrowser.fs
```diff
@@ -0,0 +1,1328 @@
...
+                                "Clone repository",
+                                (fun _ -> Browser.Dom.window.alert ("Clone " + p.web_url))
+                            )
+                        |])
+                )
+            ]
+        ]
+
+    [<ReactComponent>]
+    static member GitLabEntry() =
```
caroott, line 1167: `GitLabApi.ListUserPersonalProjects` already accepts `?userId` specifically to skip the extra `/user` lookup, but `GitLabEntry` ignores the `currentUser` it already has and never passes that ID. Every page/search/sort in "Your Repos" therefore does an avoidable extra request before the actual projects call.

src/Components/src/GenericComponents/Actionbar.fs
```diff
@@ -126,37 +162,63 @@ type Actionbar =
...
+            ?buttonSize: DaisyuiSize,
+            ?tooltipPosition: DaisyuiTooltipPosition,
+            ?buttonClassName: string
+        ) =
 
         let debug = defaultArg debug false
+        let buttonSize = defaultArg buttonSize DaisyuiSize.MD
+        let tooltipPosition = defaultArg tooltipPosition DaisyuiTooltipPosition.Bottom
 
         let selectedElements =
```
caroott, line 180: The new memoization only depends on `buttons`. If `maxNumber`, `buttonSize`, `tooltipPosition`, `buttonClassName`, or `debug` changes while the same button array instance is reused, `Actionbar.Main` will keep rendering the old layout/styling. That makes the shared component API subtly incorrect for dynamic callers.

src/Components/src/DataHub/DataHubSidebar.fs
```diff
@@ -0,0 +1,537 @@
...
+                                | Some el ->
+                                    let el = el :?> HTMLElement
+                                    Some(box el?dataset?filepath)
+                                | None -> None
+                            )
+                    )
+            ]
+        ]
+
+// ARC browser panel
```
Etschbeijer, line 208: Why is all this code commented out?

src/Components/src/DataHub/DataHubBrowser.fs
Freymaurer: Good point! Will fix + test

src/Components/src/DataHub/Types.fs
Freymaurer: I dont understand

Etschbeijer (commented):
lgtm

src/Components/src/DataHub/Types.fs
caroott: - The UI sort choice `LastUpdated` is converted to `ProjectSortField.UpdatedAt` here
- That field is turned into the GitLab query `order_by=updated_at` in GitLabApi.fs
- Now each repo row displays "Updated ... ago" from `project.last_activity_at`, not from `updated_at`, in DataHubBrowser.fs

src/Components/src/DataHub/DataHubSidebar.fs
Freymaurer: because this component was just a example implementation and not fleshed out. I moved the datahub browser part to a new component and added around 500% styling and functionality. The DataHubSidebar still contains starting material for other components, but is currently not in a state that requires an entry component or a test

src/Components/src/DataHub/DataHubBrowser.fs
Freymaurer: I changes the Select organisation option to be disabled! 😅 took me some time to understand the issue you were trying to showcase

src/Components/src/DataHub/Types.fs
Freymaurer: It is converted, because GitLab does not give me a true "updated_at" field. Below you can find the response i get from gitlab for an example ARC. As you can see GitLab only gives me `last_activity_at` and `created_at` for date time. 

```json

{
    "id": 3873,
    "description": null,
    "name": "FT_Genome_Biology_2026",
    "name_with_namespace": "PlantChromatin / FT_Genome_Biology_2026",
    "path": "FT_Genome_Biology_2026",
    "path_with_namespace": "plantchromatin/FT_Genome_Biology_2026",
    "created_at": "2026-03-19T23:37:21.314Z",
    "default_branch": "main",
    "tag_list": [],
    "topics": [],
    "ssh_url_to_repo": "git@git.nfdi4plants.org:plantchromatin/FT_Genome_Biology_2026.git",
    "http_url_to_repo": "https://git.nfdi4plants.org/plantchromatin/FT_Genome_Biology_2026.git",
    "web_url": "https://git.nfdi4plants.org/plantchromatin/FT_Genome_Biology_2026",
    "readme_url": "https://git.nfdi4plants.org/plantchromatin/FT_Genome_Biology_2026/-/blob/main/Readme.md",
    "forks_count": 0,
    "avatar_url": null,
    "star_count": 0,
    "last_activity_at": "2026-03-23T08:03:29.035Z",
    "visibility": "public",
    "namespace": {
        "id": 1532,
        "name": "PlantChromatin",
        "path": "plantchromatin",
        "kind": "group",
        "full_path": "plantchromatin",
        "parent_id": null,
        "avatar_url": null,
        "web_url": "https://git.nfdi4plants.org/groups/plantchromatin"
    }
}
```

src/Components/src/DataHub/DataHubBrowser.fs
Freymaurer: Changed it!

src/Components/src/GenericComponents/Actionbar.fs
Freymaurer: Good catch!

Etschbeijer (approved):
lgtm

PR #1020, opened 2026-03-20
move GitValidation tests to fable compatible suite

caroott:
Move the git tests from the dotnet only test suite to the fable compatible test suite


Freymaurer:
Are you teting actual git calls or mocks?

caroott:
> Are you teting actual git calls or mocks?

mocks, the actual calls on a local test are already in

PR #1021, opened 2026-03-24
Epic/store data/electron code structure refactor

Freymaurer:
Done!

Issues:

- [x] some components used relative paths, some absolute paths. I changed a lot to relative so we can sue ARCtrl logic to extract functions targeting ARCfiles. This means clicking on FileTree does not create the correct path.


Etschbeijer (approved):
Lgtm

PR #1024, opened 2026-03-30
Make GitMergeConflictViewer conflicts persistent

caroott:
Previously, merge conflicts resolved via UI would no longer be displayed as soon as manual edits happened due to line indexing changes. 

This PR makes merge-conflict resolution state in the Git merge viewer persistent across edits and controlled-content updates.

It introduces a session model that tracks each conflict entry with stable IDs, reconciles incremental text edits (instead of resetting state), and keeps valid Undo actions available after manual edits. If a manual edit crosses a resolved conflict boundary, only that entry’s Undo is disabled with a clear “Undo unavailable” reason. Full-document replacements reset the session cleanly.


Etschbeijer (approved):
lgtm

PR #1028, opened 2026-04-07
Git UI / DataHub browser

caroott:
This PR introduces a full Git workflow UI and main-process Git service layer for the Electron app, and integrates the DataHub browser / auth work from `electron-datahub-view`.

---

### Git sidebar
Adds a collapsible left-sidebar Git panel, which shows current branch, ahead/behind counts, changed files with status codes, and a commit composer supporting both selective and full-commit modes.

image

### Git fetch / pull / push / sync
Wires fetch, pull, push, and sync (publish-then-pull) operations through typed IPC to `GitService`. All remote operations report real-time progress to the renderer via `gitProgressUpdate`.

### Git LFS integration
- Automatic large-file tracking at a configurable MB threshold during commit
- `Download Large Files` toggle (persisted in repo git config) controls LFS smudge on pull and clone
- One-click `git lfs install` retry flow when LFS is not installed
- Outbound push planning: dry-run detects LFS pointers before pushing and uploads them first

### Git diff viewer
Clicking a changed file opens a full-screen side-by-side word-diff viewer

image

image

### Merge conflict resolution
Detects in-progress merges from git status and routes conflicted files to `GitMergeConflictViewer`. After resolving each file the renderer auto-advances to the next conflict. When the last conflict is resolved the merge is committed automatically.

### Branch management
Switch branch and create branch from the sidebar. Checkout triggers a file-tree refresh.

### Renderer Git state machine
Shared `GitStateCtx` manages all Git state with request-ID race-condition protection, LFS install retry semantics, and page-load lifecycle.

### Main-process Git services
`GitService`, `GitProvisioningService`, `GitLfsService`, `GitLfsAdapter`, and `GitAuthAdapter` implement the full back-end:
- Auth token injection per remote host via `GitTokenProvider`
- Clone with authenticated-then-unauthenticated fallback and optional post-clone LFS pull

### IPC contracts
New `IGitApi` interface fully registered in main, preload, and renderer API layers.

### Tests
- `GitService.test.fs`: local workflow, LFS push planning, pathspec security, diff correctness
- `GitValidation.test.fs`: failure classification, branch/remote validation, SimpleGit binding surface
- `GitWorkflow.test.fs` (new suite): renderer state machine — merge deduplication, stale response filtering, sync flow, LFS install retry

---

### DataHub browser
Full repository explorer (`DataHubBrowser.ExplorePanel`) with four tabs: All ARCs, Most Starred, Your Repos, Your Organisations. Supports paginated browsing, keyword search, and group/org drill-down.

### Clone from DataHub
Download button on each repository card picks a local destination, clones via the Git clone IPC (respecting the LFS download preference), and opens the cloned ARC automatically.

### Multi-account auth
Sign in to multiple DataHub instances simultaneously. Accounts are persisted with Electron `safeStorage`-encrypted PATs. Active account can be switched or removed from the account manager UI, with cross-window broadcast of auth state changes.

### Auth DTO unification
`AuthStateDto`, `AccountSummary`, and `AuthUserDto` are defined once in `Swate.Components.Authentication.Types` and shared across the Components, Shared, and Main layers — no duplicate definitions.

### Auth state context
`AuthStateCtx` provider propagates auth state to all renderer components. `Navbar` renders the active user avatar and account manager. DataHub tabs requiring authentication are gated on `authCtx.ActiveAccount`.

### GitLab browsing IPC
`IGitLabApi` bridges `loadAllRepos`, `loadMostStarredRepos`, `loadUserRepos`, `loadOrganisationGroups`, and `loadOrganisationRepos` from the renderer to the main-process `GitLabApi` client.


src/Electron/src/Renderer/Components/MainContent/ArcTarget.fs
```diff
@@ -34,7 +38,11 @@ let ArcTarget () =
                 | PageState.LandingDraftPage -> LandingDraftTarget()
                 | PageState.NotesDraftPage -> NotesDraftTarget()
                 | PageState.NotesSearchPage -> NotesSearchTarget()
+                | PageState.GitDiffPage diffData -> GitDiffTarget.Main diffData
+                | PageState.GitMergeConflictPage mergeData -> GitMergeConflictTarget.Main mergeData
+                | PageState.GitUnsupportedPage unsupportedPage -> GitUnsupportedTarget.Main unsupportedPage
+                | PageState.DataHubBrowser -> DataHubBrowserTarget()
```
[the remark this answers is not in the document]
Freymaurer: To explain why i Added Datahub target before ArcRootPath match case:
You might want to browse datahub before opening an ARC. So we must match for datahub page before matching for ArcRootPath.

To better show this logic, i unified InitTarget.fs and ArcTarget.fs into Main.fs

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -0,0 +1,1174 @@
...
+        |> String.concat " | "
+
+    let formatThresholdInput (thresholdMb: int) = string thresholdMb
+
+[<Erase; Mangle(false)>]
+type GitSidebar =
+
+    [<ReactComponent>]
+    static member OperationStatusNotice
+        (?currentProgress: GitSidebarProgress, ?busyNotice: string, ?errorNotice: string, ?busyTestId: string, ?errorTestId: string)
```
Freymaurer: It seems like we cannot have both GitSidebarProgress and busyNotice at the same time. Does it make sense to refactor it to:

```
[<RequireQualifiedAccess>]
type GitRunStatus =
| Busy of string
| Progress of string
| Idle // this or keep it as option
```

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -0,0 +1,1174 @@
...
+            ?errorNotice: string
+        ) =
+
+        let busyNotice = busyNotice
+        let errorNotice = errorNotice
+        let selectedFile = selectedFile
+
+        let localError, setLocalError = React.useState (None: string option)
+        let activeAction, setActiveAction = React.useState (None: string option)
+        let isAdvancedActionsOpen, setIsAdvancedActionsOpen = React.useState false
```
Freymaurer: multiple modals/dialogs can be merged into a single state using DU. At least if they are not meant to overlap

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -0,0 +1,1174 @@
...
+            ]
+        ]
+
+    [<ReactComponent>]
+    static member Main
+        (
+            status: GitSidebarStatus,
+            changedFiles: GitSidebarChange[],
+            branchOptions: GitSidebarBranchOption[],
+            onRefresh: unit -> JS.Promise<Result<unit, string>>,
```
Freymaurer: This is nitpicking, if you feel like it you can change it:

Having these many non-optional parameters with similiar types, MAY be unified into a (ano) record to improve type safety if we move around params/remove params or add more.

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -0,0 +1,1174 @@
...
+                    current
+                    |> Set.filter (fun path ->
+                        changedFiles
+                        |> Array.exists (fun change -> String.Equals(change.Path, path, StringComparison.Ordinal))
+                    )
+                )),
+            [| box changedFiles |]
+        )
+
+        React.useEffect (
```
Freymaurer: This is used to refresh local input? Maybe add a comment, as useEffects can be problematic so it is good to make intent clear

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -0,0 +1,1174 @@
...
+            (fun () -> setDownloadLargeFilesInput downloadLargeFiles),
+            [| box downloadLargeFiles |]
+        )
+
+        React.useEffect (
```
Freymaurer, line 1061: same as `setDownloadLargeFilesInput`

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -0,0 +1,1204 @@
...
+            && not (String.IsNullOrWhiteSpace normalizedLfsThresholdInput)
+            && not (
+                String.Equals(
+                    normalizedLfsThresholdInput,
+                    GitSidebarInternal.formatThresholdInput lfsAutoTrackThresholdMb,
+                    StringComparison.Ordinal
+                )
+            )
+
+        Html.div [
```
Freymaurer, line 1246: Starting here this is an insanely large component. Maybe you can split it into some sub-components. You can keep them in this file, just to make the easier to read.

src/Electron/src/Main/Git/GitlfsAdapter.fs
```diff
@@ -50,102 +114,128 @@ type NodeGitLfsAdapter() =
...
+                                ]
+
+                            let proc: obj =
+                                childProcessDynamic?spawn ("git", (args |> List.toArray), spawnOptions)
+
+                            let mutable output = ""
+                            let mutable errorOut = ""
+                            let mutable finished = false
+
+                            proc?stdout?on (
```
Freymaurer, line 263: Does it make sense to create actual `proc` fable bindings, to improve type safety or should we do this, as soon as something breaks apart?

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -0,0 +1,451 @@
...
+    | SkipLfsUpload
+    | UploadLfsObjects of refSpec: string
+
+let private lfsPointerOidPattern = Regex("^oid sha256:[0-9a-f]{64}$", RegexOptions.Multiline)
+let private lfsPointerSizePattern = Regex("^size \\d+$", RegexOptions.Multiline)
+let private maxLfsPointerProbeBytes = 1024L
+let private childProcessDynamic: obj = importAll "node:child_process"
+let mutable private cachedSystemInstalled = false
+
+let private tryExecGitText
```
Freymaurer: This is a duplicate, right? I am sure i already saw this function somewhere.

src/Electron/src/Renderer/Components/MainContent/ArcTarget.fs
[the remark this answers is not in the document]
caroott: restored it: https://github.com/nfdi4plants/Swate/pull/1028/changes/458e1055e1dc121c3a6d446ff047431b5e718855

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -0,0 +1,451 @@
...
+
+let private isLfsPointerContent (content: string) =
+    let normalized =
+        content.Replace("\r\n", "\n").Trim()
+
+    normalized.StartsWith("version https://git-lfs.github.com/spec/v1", StringComparison.Ordinal)
+    && lfsPointerOidPattern.IsMatch normalized
+    && lfsPointerSizePattern.IsMatch normalized
+
+let private containsOutboundLfsPointerObjects
```
Freymaurer: Can you please explain what this function is intended to do?

src/Electron/src/Main/IPC/IGitApi.fs
```diff
@@ -222,6 +340,21 @@ let api: IGitApi = {
...
+                let! result =
+                    GitService.setLfsSettings
+                        arcPath
+                        {
+                            AutoTrackThresholdMb = settings.AutoTrackThresholdMb
+                            DownloadLargeFiles = settings.DownloadLargeFiles
+                        }
+
+                return toGitOperationResult (fun () -> Some "Git LFS threshold updated.") None result
+        }
```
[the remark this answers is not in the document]
caroott: https://github.com/nfdi4plants/Swate/pull/1028/changes/724d5d97000e060d378c845b4bfa0f58e65e92af

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -0,0 +1,451 @@
...
+    (objectIds: string[])
+    (git: ISimpleGit)
+    : JS.Promise<Result<bool, 'Failure>> =
+    promise {
+        let mutable hasLfsPointer = false
+        let mutable failure: 'Failure option = None
+
+        for objectId in objectIds do
+            if not hasLfsPointer && failure.IsNone then
+                let! objectTypeResult = runSimpleGitRaw (fun currentGit -> currentGit.raw [| "cat-file"; "-t"; objectId |]) git
```
Freymaurer, line 484: What is the exact reason why you use `git cat-file` to check file size instead of node `fs.statSync`?

src/Electron/src/Main/Git/GitProvisioningService.fs
```diff
@@ -346,6 +348,17 @@ let private cloneWithGit
         })
         git
 
+let private applyCloneSkipSmudge (git: ISimpleGit) =
```
Freymaurer, line 241: This sets an env variable. Can you add a comment for your reasoning, as this is a rather important thing

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -24,19 +30,52 @@ type GitFailure = {
 
 type GitResult<'T> = Result<'T, GitFailure>
 
+[<RequireQualifiedAccess>]
+type GitBranchRefKind =
```
Freymaurer: I think i saw that type in GitSidebar types

src/Electron/src/Main/Git/GitService.fs
Freymaurer, line 1: Please check some of the types here, i think they are defined in GitSidebar types too.

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -46,15 +85,48 @@ type GitDiffSummaryDto = {
...
+    UpdatedStatus: GitStatusDto
+    RemainingConflictedPaths: string[]
+    NextConflictedPath: string option
+}
+
+type GitPullResult = { Warning: GitFailure option }
+
 type GitProgressCallback = GitInternals.GitProgressCallback
 
+let private fsPromisesDynamic: obj = importAll "fs/promises"
```
Freymaurer: Please move bindings to a specific bindings file

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -208,9 +379,638 @@ let private validateOptionalBranchName (branchName: string option) =
...
+
+[<Emit("$0.toString('utf8')")>]
+let private bufferToUtf8String (buffer: obj) : string = jsNative
+
+let private readFileBufferAsync (absolutePath: string) : JS.Promise<obj> =
+    fsPromisesDynamic?readFile (absolutePath)
+    |> unbox<JS.Promise<obj>>
+
+let private explicitlyUnsupportedExtensions =
+    set [
```
Freymaurer, line 337: for merge?

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -208,9 +379,638 @@ let private validateOptionalBranchName (branchName: string option) =
...
+            | None ->
+                return Error(createCommitLfsValidationFailure thresholdMb (oversizedPaths.ToArray()))
+    }
+
+// GitService keeps the pull preference application here because it changes how the authenticated pull workflow checks out the working tree.
+let private applyLfsDownloadPreference (downloadLargeFiles: bool) (git: ISimpleGit) =
+    if downloadLargeFiles then
+        git
+    else
+        git.env ("GIT_LFS_SKIP_SMUDGE", "1")
```
Freymaurer: There is another instance, in which we set this env variable. Please use a String literal for the key

src/Electron/src/Main/IPC/IGitApi.fs
```diff
@@ -58,12 +67,64 @@ let private toStatusDto (status: GitService.GitStatusDto) : GitStatusDto = {
         })
 }
 
+let private toBranchKind (kind: GitService.GitBranchRefKind) =
```
Freymaurer: these converters look like code duplication wrappers. As mentioned above, please do not add type duplications.

src/Electron/src/Renderer/Components/MainContent/DataHubBrowserTarget.fs
```diff
@@ -0,0 +1,122 @@
...
+                    | Ok _ -> closeBrowser ()
+                    | Error error ->
+                        Browser.Dom.console.error ($"[Swate] Could not open cloned ARC: {error.Message}")
+        }
+        |> Promise.start
+
+    Html.div [
+        prop.className "swt:size-full swt:flex swt:flex-col"
+        prop.children [
+            GitSidebar.OperationStatusNotice(
```
Freymaurer, line 96: I think it would be awesome to have a unified "Toast" ( https://storybooks.fluentui.dev/react/?path=/docs/components-toast--docs#default ) to display git status

These is a larger change, as it would make sense to have a unified TOAST provider component, which can be fed the git status as well as possibly others.

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+[<Hook>]
+let useGitState () = React.useContext GitStateCtx
+
+[<ReactComponent>]
+let GitStateCtxProvider (children: ReactElement) =
+
+    let appStateCtx = Renderer.Context.AppStateCtx.useAppState ()
+    let pageStateCtx = Renderer.Context.PageStateCtx.usePageState ()
+    let gitState, elmishDispatch = React.useElmish ((fun () -> init appStateCtx.state), update, [| box appStateCtx.state |])
+    let gitStateRef = React.useRef gitState
```
Freymaurer: I do not understand this pattern? This seems to hotfix some underlying issue.

`useElmish` hook uses React.useSyncExternalStore under the hood, so it should be rerender safe?

In addition it seems like you do not really make use of "Cmd" module provided by Elmish (e.g. to call fetch and directly handle results inside update.

If you do not intend to use the full capacity of elmish Cmd module, you might want to change it to reacts "useReducer" (same state behavior without possibility for promise/async call management)


No matter how you decide. Please to not store the useElmish state inside a react ref. This is bad practise and MUST be avoided

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+
+[<Hook>]
+let useGitState () = React.useContext GitStateCtx
+
+[<ReactComponent>]
+let GitStateCtxProvider (children: ReactElement) =
+
+    let appStateCtx = Renderer.Context.AppStateCtx.useAppState ()
+    let pageStateCtx = Renderer.Context.PageStateCtx.usePageState ()
+    let gitState, elmishDispatch = React.useElmish ((fun () -> init appStateCtx.state), update, [| box appStateCtx.state |])
```
Freymaurer: will have to check init, update when i find them

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+    let gitStateRef = React.useRef gitState
+
+    gitStateRef.current <- gitState
+
+    let dispatch msg =
```
Freymaurer: yeah, again, remove all traces of the elmish-state -> ref pattern

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+        let nextState = transition msg gitStateRef.current
+        gitStateRef.current <- nextState
+        elmishDispatch msg
+
+    let getState () = gitStateRef.current
+    let isArcLoaded () = appStateCtx.state.IsSome
+
+    let refresh () = promise {
```
Freymaurer: This is an example of you, not using elmish to its potential. Inside elmish you can use `Cmd.ofPromise.either` to make promise calls with robust state handling and directly map responses to a new elmish message

Freymaurer: So this function is bad "useElmish" pattern

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+            dependencies
+            isArcLoaded
+            getState
+            dispatch
+            GitBusyOperation.CommittingSelectedChanges
+            request.Paths
+            true
+            request.Message
+
+    let commitAll (message: string) =
```
Freymaurer: This is a function call based on state, it should be called using a simple Elmish Msg, while having the core logic inside a helper function and the elmish `update` function

Freymaurer: same for almost all functions inside this component

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+        Renderer.Context.GitWorkflow.confirmMergeResolution dependencies isArcLoaded getState dispatch request
+
+    let ipcHandler: IMainUpdateRendererApi = {
+        IMainUpdateRendererApi.empty with
+            gitProgressUpdate =
+                fun progress ->
+                    dispatch (SetCurrentProgress(Some(mapProgress progress)))
+    }
+
+    React.useEffectOnce (fun _ -> Remoting.init |> Remoting.buildHandler ipcHandler)
```
Freymaurer: `useElmish` above actually has a overload using a "subscription":

https://github.com/fable-hub/Feliz/pull/684

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,342 @@
...
+
+    React.useEffect (
```
Freymaurer: This seems to be necessary to handle ambiguity between PageState and GitPage. Remove GitPage and instead only rely on PageState. Not having a singular ground truth is a bad pattern

src/Electron/src/Renderer/context/GitWorkflow.fs
Freymaurer, line 1: Image

Let me trust you on this one

src/Electron/src/Renderer/GitApiClient.fs
Freymaurer, line 1: Can you tell me your reasoning why you created a helper function for every member of the gitApi? This seems like boilerplate that reduced readability?

src/Electron/src/Swate.Electron.Shared/GitTypes.fs
```diff
@@ -31,21 +33,59 @@ type GitFailureKind =
     | Network
     | Timeout
     | Canceled
+    | LfsInstallRequired
     | Unknown
 
+[<StringEnum(CaseRules.None)>]
+type GitBranchRefKind =
```
Freymaurer, line 42: I think this is the 3rd time i see this type... And some of the types below too

Freymaurer (changes requested):

src/Components/src/GitSidebar/GitSidebar.fs
caroott: good point, they are mutually exclusive. will refactor to a DU like you suggested

src/Components/src/GitSidebar/GitSidebar.fs
caroott: they dont overlap, makes sense to unify them into a single DU. will change

src/Components/src/GitSidebar/GitSidebar.fs
caroott: yeah i think a record makes sense here, especially since most params share the same type signature

src/Components/src/GitSidebar/GitSidebar.fs
caroott: yes it syncs the local input state when the prop changes

src/Components/src/GitSidebar/GitSidebar.fs
caroott: agreed, will split the render logic into smaller sub-components within the same file

src/Electron/src/Main/Git/GitlfsAdapter.fs
caroott: for now the dynamic interop works fine and the usage is limited. maybe add proper bindings when we extend the usage or something breaks?

src/Electron/src/Main/Git/GitLfsService.fs
caroott: yes, this exists in GitLfsAdapter already. Saw it as well and wanted to fix it with the other issues

src/Electron/src/Main/Git/GitLfsService.fs
caroott: it checks whether the outbound objects that a git push would send contain any LFS pointer files. this is needed to decide if we need to run git lfs push before the actual push

src/Electron/src/Main/Git/GitLfsService.fs
caroott: git cat-file -s gives the size of the staged blob in the git object store, not the working tree file. this matters because the file on disk could differ from what's staged

src/Electron/src/Main/Git/GitService.fs
caroott: this is used to read file contents for the merge conflict viewer. we need to detect whether a conflicted file is binary or text

src/Electron/src/Renderer/context/GitStateCtx.fs
caroott: init and update are defined at the top of the same file

src/Electron/src/Renderer/GitApiClient.fs
caroott: the wrappers map the IPC Result<'T, GitFailure> to Result<'T, string> for the UI layer, so the components don't need to know about GitFailure. but you're right it's a lot of boilerplate. i could simplify with a generic mapper and use the api members directly with it

src/Electron/src/Main/Git/GitLfsService.fs
Freymaurer: please add a comment :)

src/Electron/src/Main/Git/GitLfsService.fs
Freymaurer: If you are only checking if the file in question is a pointer file, you might get more robust/quicker results if you use fs.stats.

I also had to do something similiar for arc-export, where we ran into system-out-of-memory exceptions by running regex over multi-gb files, trying to match regex patterns on the potential pointer files. Please make sure your functions is robust to check actual file size before running any regex on the content:

See my very quick hotfix implementation: https://github.com/nfdi4plants/arc-export/blob/9283a6e64ade0a5dc8536b5fcd63c3795cd7e5c1/src/arc-export/GitLFS.fs#L12

src/Electron/src/Renderer/context/GitStateCtx.fs
Freymaurer: 😞 ups

Freymaurer: wait, no! Where? There is no init above?

```fsharp
module Renderer.Context.GitStateCtx

open Browser.Dom
open Fable.Core
open Feliz
open Feliz.UseElmish

open Renderer.Types
open Swate.Components.GitSidebarTypes
open Swate.Electron.Shared
open Swate.Electron.Shared.GitTypes
open Swate.Electron.Shared.IPCTypes

open Renderer.Context.GitWorkflow

type GitStateController = {
    state: GitState
    refresh: unit -> JS.Promise<Result<unit, string>>
    fetch: unit -> JS.Promise<Result<unit, string>>
    pull: unit -> JS.Promise<Result<unit, string>>
    push: unit -> JS.Promise<Result<unit, string>>
    cloneRepository: GitCloneRepositoryRequest -> JS.Promise<Result<string, string>>
    sync: unit -> JS.Promise<Result<unit, string>>
    commitSelection: GitSidebarCommitSelectionRequest -> JS.Promise<Result<unit, string>>
    commitAll: string -> JS.Promise<Result<unit, string>>
    saveLfsAutoTrackThreshold: int -> JS.Promise<Result<unit, string>>
    saveDownloadLargeFiles: bool -> JS.Promise<Result<unit, string>>
    createBranch: GitSidebarCreateBranchRequest -> JS.Promise<Result<unit, string>>
    switchBranch: string -> JS.Promise<Result<unit, string>>
    selectChange: GitSidebarChange -> JS.Promise<Result<unit, string>>
    confirmMergeResolution: GitConfirmMergeResolutionRequest -> JS.Promise<Result<unit, string>>
}

let private mapDiffPageResult (_requestedPath: string) = function
    | Ok(GitPageLoadResultDto.Loaded diffData) -> Ok(GitPage.Diff diffData)
    | Ok(GitPageLoadResultDto.Unsupported unsupportedPage) -> Ok(GitPage.Unsupported unsupportedPage)
    | Error message -> Error message

let private mapMergeConflictPageResult (_requestedPath: string) = function
    | Ok(GitPageLoadResultDto.Loaded mergeData) -> Ok(GitPage.MergeConflict mergeData)
    | Ok(GitPageLoadResultDto.Unsupported unsupportedPage) -> Ok(GitPage.Unsupported unsupportedPage)
    | Error message -> Error message

let private dependencies: GitDependencies = {
    getGitStatus = Renderer.GitApiClient.getGitStatus
    getGitBranches = Renderer.GitApiClient.getGitBranches
    getGitLfsSettings = Renderer.GitApiClient.getGitLfsSettings
    loadDiffPage =
        fun requestedPath ->
            promise {
                let! result = Renderer.GitApiClient.getGitDiffViewData requestedPath
                return mapDiffPageResult requestedPath result
            }
    loadMergeConflictPage =
        fun requestedPath ->
            promise {
                let! result = Renderer.GitApiClient.getGitMergeConflictViewData requestedPath
                return mapMergeConflictPageResult requestedPath result
            }
    installGitLfs = Renderer.GitApiClient.installGitLfs
    gitFetch = Renderer.GitApiClient.gitFetch
    gitPull = Renderer.GitApiClient.gitPull
    gitPush = Renderer.GitApiClient.gitPush
    gitCloneRepository = Renderer.GitApiClient.gitCloneRepository
    createBranch = Renderer.GitApiClient.createBranch
    checkoutBranch = Renderer.GitApiClient.checkoutBranch
    gitStagePaths = Renderer.GitApiClient.gitStagePaths
    gitUnstagePaths = Renderer.GitApiClient.gitUnstagePaths
    gitCommit = Renderer.GitApiClient.gitCommit
    setGitLfsSettings = Renderer.GitApiClient.setGitLfsSettings
    confirmGitMergeResolution = Renderer.GitApiClient.confirmGitMergeResolution
    confirmInstall = fun message -> window.confirm message
}

let GitStateCtx =
    React.createContext<GitStateController> (
        {
            state = GitState.Empty
            refresh = fun () -> promise { return Ok() }
            fetch = fun () -> promise { return Ok() }
            pull = fun () -> promise { return Ok() }
            push = fun () -> promise { return Ok() }
            cloneRepository = fun _ -> promise { return Ok "" }
            sync = fun () -> promise { return Ok() }
            commitSelection = fun _ -> promise { return Ok() }
            commitAll = fun _ -> promise { return Ok() }
            saveLfsAutoTrackThreshold = fun _ -> promise { return Ok() }
            saveDownloadLargeFiles = fun _ -> promise { return Ok() }
            createBranch = fun _ -> promise { return Ok() }
            switchBranch = fun _ -> promise { return Ok() }
            selectChange = fun _ -> promise { return Ok() }
            confirmMergeResolution = fun _ -> promise { return Ok() }
        }
    )

[<Hook>]
let useGitState () = React.useContext GitStateCtx

[<ReactComponent>]
let GitStateCtxProvider (children: ReactElement) =

    let appStateCtx = Renderer.Context.AppStateCtx.useAppState ()
    let pageStateCtx = Renderer.Context.PageStateCtx.usePageState ()
    let gitState, elmishDispatch = React.useElmish ((fun () -> init appStateCtx.state), update, [| box appStateCtx.state |])
```

src/Electron/src/Renderer/GitApiClient.fs
Freymaurer: Might make sense to do it that way. You could even add a member to GitFailure to transform the message to string or exn. So it would be very findable api

src/Electron/src/Renderer/context/GitStateCtx.fs
caroott: here by now: https://github.com/nfdi4plants/Swate/pull/1028/changes#diff-0342f8959cea9727fa02fa67e76f48fc2c6366dc224e377415c5f46a6ce0ce5cR372

caroott:
- The renderer Git flow was reworked into a real Elmish `Msg`/`Cmd` pipeline: request payloads are snapshotted from model state, refresh/page-load/write orchestration moved out of the old imperative ref pattern, `GitStateCtx` is now a thin `useElmish` bridge, and refresh failures clear stale page state again.
- ARC-switch and async race handling were tightened by tagging long-running work with ARC sessions, ignoring late completions from a previous ARC, and stabilizing the Git LFS install-retry path.
- Git LFS transport/auth was tightened on both clone and push: clone auth is now scoped to the GitLab host instead of a global header, the old unauthenticated hydration retry was removed, and outbound LFS push now plans from remote truth and uploads exact object IDs via spawned git helpers.
- Renderer/Core tests were updated around the new APIs and expanded for auth scoping, exact LFS uploads, stale-session handling, and the longer-running LFS integration cases.

src/Electron/src/Renderer/context/GitWorkflow.fs
```diff
@@ -0,0 +1,1223 @@
...
+    PathsToCommit = state.ChangedFiles |> Array.map _.Path |> distinctPaths
+    CurrentlyStagedPaths = [||]
+}
+
+let buildUpdatedLfsSettings (state: GitState) (thresholdMb: int option) (downloadLargeFiles: bool option) = {
+    AutoTrackThresholdMb = thresholdMb |> Option.defaultValue state.LfsAutoTrackThresholdMb
+    DownloadLargeFiles = downloadLargeFiles |> Option.defaultValue state.DownloadLargeFiles
+}
+
+let private effectCmd (effect: unit -> unit) : Cmd<'msg> = [ fun _dispatch -> effect () ]
```
caroott: use Cmd.ofFunc or Cmd.ofEffect

src/Electron/src/Renderer/context/GitStateCtx.fs
```diff
@@ -0,0 +1,188 @@
...
+    commitAll: string -> JS.Promise<Result<unit, string>>
+    saveLfsAutoTrackThreshold: int -> JS.Promise<Result<unit, string>>
+    saveDownloadLargeFiles: bool -> JS.Promise<Result<unit, string>>
+    createBranch: GitSidebarCreateBranchRequest -> JS.Promise<Result<unit, string>>
+    switchBranch: string -> JS.Promise<Result<unit, string>>
+    selectChange: GitSidebarChange -> JS.Promise<Result<unit, string>>
+    confirmMergeResolution: GitConfirmMergeResolutionRequest -> JS.Promise<Result<unit, string>>
+}
+
+let private dispatchPromise
```
caroott: Promise.either instead of dispatchPromise. Is this needed?

src/Electron/src/Renderer/context/GitWorkflow.fs
```diff
@@ -0,0 +1,1223 @@
...
+        return! runCommitAttemptAsync deps prepared
+    | SaveLfsSettings(busyOperation, settings, _) ->
+        return! runSaveLfsSettingsAttemptAsync deps busyOperation settings
+    | CreateBranch(request, _) ->
+        return! runSimpleWriteAttemptAsync deps GitBusyOperation.CreatingBranch (fun () -> deps.createBranch request)
+    | SwitchBranch(request, _) ->
+        return! runSimpleWriteAttemptAsync deps GitBusyOperation.SwitchingBranch (fun () -> deps.checkoutBranch request)
+}
+
+let transition (msg: Msg) (model: GitState) =
```
caroott: transition makes updates by creating a copy each time
no double matching on msg. Matches msg where transition is called and again in transition.
Use {model with RefreshState = GitRefreshState.Idle ...} and so on instead or create a member if it happens more often.

src/Electron/src/Renderer/context/GitWorkflow.fs
```diff
@@ -0,0 +1,1223 @@
...
+            match arcPath with
+            | Some _ ->
+                Cmd.batch [
+                    applyPageChangeCmd setPageState GitPageChange.Clear
+                    Cmd.ofMsg (RefreshRequested ignoreReply)
+                ]
+            | None -> applyPageChangeCmd setPageState GitPageChange.Clear
+
+        nextModel, cmd
+    | RefreshRequested reply when model.CurrentArcPath.IsNone ->
```
caroott: does not really require "reply", as it is not used

Issue #1031, opened 2026-04-10
[Electron] Git UI Feedback


Freymaurer:
We are missing some edge case handling:

- [x] Do not show git ui if not open in ARC, see vs code as example:

<img width="327" height="303" alt="Image" src="https://github.com/user-attachments/assets/1e1fea56-6848-4f3e-95f5-0932af5a5284" />

- [x] Then when ARC is opened, but git is missing, do not show ui, instead show info text + button to init git. This should also include a name field for remote and then add it accordingly, using the selected datahub. Everything related to remote should only be available when the user is logged in. If they are not logged in, these fields should be disabled with a WARNING-color info text below.

```bash
git remote add origin <REMOTE_URL>
```

<img width="405" height="220" alt="Image" src="https://github.com/user-attachments/assets/207bbcaf-444f-49e8-83b2-5e3bb97c96bd" />

- [x] "Changes" list is not virtualized and creates huge performance issue when the list goes into hundreds. We have virtualisation logic in `src/Components/src/Util/Virtual.fs`. If you need examples you can check the Table component or the TermSearch component. Feel free to ask me on this topic.

<img width="354" height="402" alt="Image" src="https://github.com/user-attachments/assets/5100b840-d46d-49b6-8ada-25fae2ae5c8e" />

- [x] When opening the git view with changes i get these warnings in my console. These could hint at a underlying issue. 

<img width="289" height="863" alt="Image" src="https://github.com/user-attachments/assets/22a42e51-021b-4ccc-8c89-7f490cf3c297" />

- [x] Seems to register empty changes? for multiple files the diff view is empty

<img width="1457" height="844" alt="Image" src="https://github.com/user-attachments/assets/a610beea-c405-4719-9be9-d96d213edb60" />

- [ ] UI feedback: What is the small "D." in front of the changes path name? Can we do colors for the git change type? Like "green" for added, "yellow" for changes, "red" for deleted, or something similiar to provide a quicker overview? In addition, i think the current design contains a lot of whitespace, which makes a quick overview difficult. I think a more minimal list approach might be beneficial, see vs code as example.

<img width="334" height="78" alt="Image" src="https://github.com/user-attachments/assets/db913808-6087-44d4-a588-339d29497834" />

Freymaurer:
- [x] :bug: list virtualisation is broken, as you can see in the gif below. All 1k buttons are already rendered, instead of only the ones visible + virtualisation overflow.

[...]

PR #1033, opened 2026-04-10
Git Feature Requests #1031

caroott:
This PR addresses the requests from https://github.com/nfdi4plants/Swate/issues/1031

It adds renderer-side repository-availability state and a local `git init` flow (with optional DataHub project creation + `origin` wiring), disables remote actions when not authenticated, virtualizes the changed-files list, and replaces the generic change badge with explicit labels. Includes IPC/Service additions for `gitAddRemote` and GitLab project creation and adds tests covering the new flows.

> When opening the git view with changes i get these warnings in my console. These could hint at a underlying issue.

This issue could not be reproduced. This should have been fixed already.

- simple-git returns ONE entry per path in status.files. Each file has both index and working_dir status codes consolidated into a single entry
- React keys use change.Path which is already unique since mapChanges maps status.Files 1:1 without concatenation
- No console.warn statements exist in any git-related component or service code
- The checkbox is properly controlled (prop.isChecked + prop.onChange), ruling out controlled/uncontrolled warnings
- The code uses Array.distinct on paths in staging operations, but the status.files array itself doesn't produce them

>  Seems to register empty changes? for multiple files the diff view is empty

This is taken directly from the git output. If the git output somehow registers changes on empty files they will show up. This also happens in VSCode for example. We could guard against empty diffs, but this could also suggest that something changed the file in a way we can't display.

**Key changes**

- Renderer: new repository-availability state, InitRepository flow, and empty-state UI for no-ARC / missing-repository.
- Remote bootstrap: optional creation of a DataHub/GitLab project and `git remote add origin` during init when authenticated and a project name is provided.
- UI: auth-aware gating for fetch/pull/push/sync; explicit change status labels (Added/Modified/Deleted/Renamed/Untracked/Conflict); virtualization of long changed-file lists.
- IPC & Main: new IPC contract and service handler for `gitAddRemote`, GitLab project-creation handler.
- Tests: characterization tests for the empty-diff symptom, workflow tests for missing-repo & init flows, virtualization and deleted-file render checks, and a git-service test for `addRemote`.

**Behavioral notes (what to expect)**

- Repository availability classification:
  - No ARC opened → shows Open/Download CTA.
  - ARC opened but not a git repo → shows Initialize Repository empty state (with optional DataHub project name input).
  - Repository ready → full Git sidebar with changes, branches, and actions.

- Init flow:
  - `Initialize Repository` runs local `git init`. If a DataHub project name is provided and an active account exists, the renderer will request project creation and add `origin` with the returned repository URL. Failures in remote bootstrap surface as non-blocking warnings.

- Remote action gating:
  - Fetch/pull/push/sync buttons are disabled when user is not authenticated to DataHub; a clear inline warning explains why.

- Changed-file list:
  - Large change lists are virtualized for performance (only visible rows are mounted).
  - Each change shows an explicit label and badge (Added/Modified/Deleted/Renamed/Untracked/Conflict) plus iconography, replacing the single generic "Changed" badge and raw status code.

### DIFF
--- a/src/Components/src/Api/GitLabApi.fs
+++ b/src/Components/src/Api/GitLabApi.fs
@@ -369,6 +369,46 @@ module private Internals =
                 return Error(GitLabError.NetworkError networkError)
     }
 
+    let normalizeProjectPath (name: string) =
+        System.Text.RegularExpressions.Regex("[^A-Za-z0-9._-]+")
+            .Replace(name.Trim().ToLowerInvariant(), "-")
+            .Trim('-')
+
+    let sendJson<'TResponse>
+        (url: string)
+        (pat: string)
+        (method': HttpMethod)
+        (body: obj)
+        : JS.Promise<Result<'TResponse, GitLabError>> =
+        promise {
+            if String.IsNullOrWhiteSpace pat then
+                return Error(GitLabError.InvalidRequest "Personal Access Token is required.")
+            else
+                let requestOptions = [
+                    RequestProperties.Method method'
+                    requestHeaders [
+                        HttpRequestHeaders.Custom("PRIVATE-TOKEN", pat)
+                        HttpRequestHeaders.Accept "application/json"
+                        HttpRequestHeaders.ContentType "application/json"
+                    ]
+                    RequestProperties.Body(unbox (Fable.Core.JS.JSON.stringify body))
+                ]
+
+                try
+                    let! response = fetchUnsafe url requestOptions
+
+                    if not response.Ok then
+                        return Error(mapHttpError response.Status)
+                    else
+                        try
+                            let! payload = response.json<'TResponse> ()
+                            return Ok payload
+                        with decodeError ->
+                            return Error(GitLabError.DecodeError decodeError)
+                with networkError ->
+                    return Error(GitLabError.NetworkError networkError)
+        }
+
 [<AttachMembers>]
 type GitLabApi =
 
@@ -378,6 +418,28 @@ type GitLabApi =
         let! response = Internals.sendGetSingle<GitLabCurrentUserResponse> url pat
         return response |> Result.map Internals.toCurrentUserDto
     }
+
+    static member CreateProject(baseUrl: string, pat: string, projectName: string) : JS.Promise<Result<ExploreProjectDto, GitLabError>> =
+        promise {
+            let normalizedName = projectName.Trim()
+            let normalizedPath = Internals.normalizeProjectPath normalizedName
+
+            if String.IsNullOrWhiteSpace normalizedName || String.IsNullOrWhiteSpace normalizedPath then
+                return Error(GitLabError.InvalidRequest "Repository name is required.")
+            else
+                let! response =
+                    Internals.sendJson<GitLabProjectResponse>
+                        $"{baseUrl.TrimEnd('/')}/api/v4/projects"
+                        pat
+                        HttpMethod.POST
+                        (createObj [
+                            "name" ==> normalizedName
+                            "path" ==> normalizedPath
+                            "initialize_with_readme" ==> false
+                        ])
+
+                return response |> Result.map Internals.toExploreProjectDto
+        }
 
     /// GET /api/v4/projects
     static member ListProjects
--- a/src/Components/src/Api/GitLabApi.fs
+++ b/src/Components/src/Api/GitLabApi.fs
@@ -369,11 +369,6 @@ module private Internals =
                 return Error(GitLabError.NetworkError networkError)
     }
 
-    let normalizeProjectPath (name: string) =
-        System.Text.RegularExpressions.Regex("[^A-Za-z0-9._-]+")
-            .Replace(name.Trim().ToLowerInvariant(), "-")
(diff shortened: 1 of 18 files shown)


src/Components/src/Api/GitLabApi.fs
```diff
@@ -369,6 +369,46 @@ module private Internals =
                 return Error(GitLabError.NetworkError networkError)
     }
 
+    let normalizeProjectPath (name: string) =
```
Freymaurer: This function looks awkward. Is there not some built in functionality for this in node?

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -690,6 +747,17 @@ type GitSidebar =
 
     [<ReactComponent>]
     static member private ChangedFilesList(props: ChangedFilesListProps) =
+        let scrollContainerRef = React.useElementRef ()
+
+        let rowVirtualizer =
```
Freymaurer: I assume this is for changed files. Then it is column or list virtualisation not row 😅 in addition size seems a bit large for the list elements and i am not sure if we want this many gap. I would also increase overscan (which is the number of elements rendered which are currently not visible on screen, if i remember correctly). This ensures we get a nice smooth scrolling behavior.

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -711,31 +780,41 @@ type GitSidebar =
                             prop.text "No changed files. Your repository is in sync."
                         ]
                     else
-                        Html.div [
-                            prop.className "swt:mt-2 swt:flex swt:flex-col swt:gap-2"
-                            prop.children [
-                                for change in props.ChangedFiles do
+                        let renderChangeRow (isVirtualized: bool) (index: int) (top: int option) =
```
Freymaurer: Better to define this outside as reactcomponent. Why would we do virtualisation conditional? (The isBirtualized arg?)

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -776,15 +855,16 @@ type GitSidebar =
                                                                         prop.text change.Path
                                                                     ]
                                                                     Html.span [
```
Freymaurer: This badge could be a simple subcomponent

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -802,8 +882,29 @@ type GitSidebar =
                                             ]
                                         ]
                                     ]
+
+                        let renderRows () =
```
Freymaurer: Move outside as react component

src/Components/src/GitSidebar/GitSidebar.fs
```diff
@@ -802,8 +882,29 @@ type GitSidebar =
...
+                            Html.div [
+                                prop.className "swt:mt-2 swt:flex swt:flex-col swt:gap-2"
+                                prop.children (
+                                    props.ChangedFiles
+                                    |> Array.mapi (fun index _ -> renderChangeRow false index None)
+                                )
+                            ]
+
+                        if props.ChangedFiles.Length <= 24 then
```
Freymaurer: Ahh i see, so you only want to virtualize for counts > 24. Did you have performance issues for numbers below 24 with virtualisation?

Freymaurer (changes requested):

src/Components/src/GitSidebar/GitSidebar.fs
caroott: I had a feeling that yes. So i set it at 24, because that is how many files were shown on my screen without scrolling ._.

I removed it for now, since it seems to work fine. maybe it was just because there were larger files

src/Components/src/GitSidebar/GitSidebar.fs
caroott: Renamed to changedFileListVirtualizer, estimate 96→64, gap 8→4, overscan 2→8. (f66f864)

src/Components/src/Api/GitLabApi.fs
caroott: Path is now omitted from API body, letting GitLab generate it.  (95f20e8)

src/Components/src/GitSidebar/GitSidebar.fs
caroott: Extracted as GitSidebar.ChangedFileRow private ReactComponent. isVirtualized eliminated. (f66f864)

src/Components/src/GitSidebar/GitSidebar.fs
caroott: Extracted as GitSidebar.ChangeStatusBadge. (f66f864)

src/Components/src/GitSidebar/GitSidebar.fs
caroott: Removed. Single virtualized path now replaces both rendering branches. (f66f864)

Etschbeijer (approved):
lgtm

PR #1040, opened 2026-04-14
git comparison virtualization and sidebar virtualization fix

caroott:
- Virtualized shared Git comparison row rendering for `GitDiffViewer` and `GitMergeConflictViewer` so large diffs/conflict blocks mount only visible rows plus overscan.
- Added regression coverage for large diff and merge-conflict virtualization in Storybook and Electron renderer tests.
- Fixed Git sidebar changed-file virtualization in the Electron left sidebar by keeping the Git wrapper height-bounded (https://github.com/nfdi4plants/Swate/issues/1031).

- Added coverage for the left-sidebar wrapper class decision so the Git-only `swt:h-full` behavior is protected directly.
- Replaced comparison width refresh-on-render behavior with a guarded `ResizeObserver`, while keeping the window resize fallback, so virtualized comparison headers/content stay aligned when the container resizes without a window resize event.

## Follow-Up Notes

- The sidebar coverage now checks the actual wrapper class decision used by `LeftSidebar.Main`, including that Git gets `swt:h-full` and File Explorer does not.
- The comparison resize story now reproduces stale body/header width after a parent-only resize and verifies the ResizeObserver update path.
- TDD checks were run for both review findings:
  - Sidebar test failed when the Git height class was temporarily removed, then passed after restoring it.
  - Resize story failed before the ResizeObserver change with a stale width delta, then passed after the implementation.


src/Components/src/GitComparison/GitTextComparison.fs
```diff
@@ -139,40 +186,166 @@ module internal GitTextComparisonRendering =
...
-            let rows = if List.isEmpty rows then [ Rows.emptyRow () ] else rows
+            let rows = if rows.Length = 0 then [| Rows.emptyRow () |] else rows
             let theme = defaultArg theme DiffTheme
+            let rowEstimatePx = 28
+            let overscan = 8
+            let comparisonContentWidthPx, setComparisonContentWidthPx = React.useState 0
+            let headerScrollRef: IRefValue<HTMLElement option> = React.useElementRef ()
+            let bodyScrollRef: IRefValue<HTMLElement option> = React.useElementRef ()
+
+            React.useEffect (
```
Freymaurer, line 194: PLease use observer from existing dependency: https://usehooks.com/usemeasure

src/Electron/src/Renderer/Components/LeftSidebar/MainStyles.fs
```diff
@@ -0,0 +1,21 @@
+module Renderer.Components.LeftSidebar.MainStyles
+
+open Renderer.Types
```
Freymaurer: Not sure what this is meant to be? This should be unified in the LestSidebar/Main file for all components

Freymaurer (changes requested):

src/Electron/src/Renderer/Components/LeftSidebar/MainStyles.fs
caroott: i separated it in 4c921960fdceec3c6df9eed40ef87c8ef442ce8d and unified it again in c0e4ef9d3f55598079e84d544f68864c38690eb5^^

src/Components/src/GitComparison/GitTextComparison.fs
caroott: Implemented in [c315c6e](https://github.com/nfdi4plants/Swate/pull/1040/commits/c315c6e355f1bb8a4fa0e73e7cc40afd914d4fff) and [f6a3f2e](https://github.com/nfdi4plants/Swate/pull/1040/commits/f6a3f2e6f83e4c1f7956fdb6cb0876fac6fdf830)

Etschbeijer (approved):
LGTM

Etschbeijer (approved):
gtm

Issue #1038, opened 2026-04-14
[Dev] `src\Electron\src\Renderer\MainUpdateRendererBridge.fs` react anti pattern


Freymaurer:
The file uses mutable variables to persist information. We are in a React framework so we should persist our information within React states **OR** bind to the external information using syncExternalStore.

This file requires a full refactor, which should be discussed team wide.

@caroott @Etschbeijer

Freymaurer:
## Suggested solution

Splitting IPC handlers into multiples allows us to only subscribe to the relevant ones. Then we can propagate subscribed changes via React context, or simply subscribe multiple times if the information is sufficiently small.

```fsharp
module IMainUpdateRendererIPCs =
    type IPathChangeIPC = {
        subscribe: (string option -> unit) -> (unit -> unit)
    }

    type IRecentARCsUpdateIPC = {
        subscribe: (ARCPointer[] -> unit) -> (unit -> unit)
    }

    type IAuthAccountsUpdateIPC = {
        subscribe: (AuthStateDto -> unit) -> (unit -> unit)
    }

    type IFileTreeUpdateIPC = {
        subscribe: (System.Collections.Generic.Dictionary<string, FileEntry> -> unit) -> (unit -> unit)
    }

    type IGitProgressUpdateIPC = {
        subscribe: (GitProgressDto -> unit) -> (unit -> unit)
    }

```

## Current

```fsharp
/// One Way Bridge: Main -> Renderer
type IMainUpdateRendererApi = {
    pathChange: string option -> unit
    recentARCsUpdate: ARCPointer[] -> unit
    authAccountsUpdate: AuthStateDto -> unit
    fileTreeUpdate: System.Collections.Generic.Dictionary<string, FileEntry> -> unit
    gitProgressUpdate: GitProgressDto -> unit
}
with
    static member empty = {
        pathChange = ignore
        recentARCsUpdate = ignore
        authAccountsUpdate = ignore
        fileTreeUpdate = ignore
        gitProgressUpdate = ignore
    }

```

caroott:
Fixed with #1066

PR #1054, opened 2026-04-20
Electron IPC bridge refactor: per-channel types, external-store hooks, and eager registration

caroott:
This PR addresses https://github.com/nfdi4plants/Swate/issues/1038. The renderer-side IPC bridge had architectural issues.

**Monolithic type.** A single record type `IMainUpdateRendererApi` bundled five unrelated one-way Main→Renderer channels. `Fable.Electron.Remoting` derives wire-level channel names from the record type name, so every call site (`buildClient`, `buildHandler`, `buildBridge`) had to carry the full five-field type even though each site only uses one field. This made it impossible to add, rename, or version a single channel independently.

**Hand-rolled pub/sub outside React.** `MainUpdateRendererBridge.fs` maintained mutable `Dictionary<int, 'T -> unit>` subscriber maps, a pub/sub system running entirely outside React's state model. There was no way for React to reconcile IPC-driven state changes with render scheduling, and the bridge duplicated work that `useSyncExternalStore` exists to solve.

---

#### Shared types — five independent per-channel records

`IMainUpdateRendererApi` is replaced with five single-field record types in `IPCTypes.fs`:

```fsharp
type IPathChangeApi          = { pathChange:        string option -> unit }
type IRecentARCsUpdateApi    = { recentARCsUpdate:   ARCPointer[] -> unit }
type IAuthAccountsUpdateApi  = { authAccountsUpdate: AuthStateDto -> unit }
type IFileTreeUpdateApi      = { fileTreeUpdate:     Map<string, FileEntry> -> unit }
type IGitProgressUpdateApi   = { gitProgressUpdate:  GitProgressDto -> unit }
```

Each type occupies its own wire-channel namespace. Preload registers five independent `buildBridge` calls. Every Main-side call site builds a client for exactly the type it needs.

#### Renderer bridge — `IPCStore<'T>` + `useSyncExternalStore`

`MainUpdateRendererBridge.fs` is rewritten around `IPCStore<'T>`, a minimal external-store adapter:

- `GetSnapshot() : 'T voption` — returns `ValueNone` until the first IPC event, then `ValueSome value` for the latest payload. `ValueSome None` is representable (e.g. a closed-ARC path change), distinguishable from `ValueNone`.
- `Subscribe(listener) : unit -> unit` — adds a listener and returns a dispose function. Integer IDs make disposal deterministic.
- `Update(value)` — snapshots the new value and notifies all current listeners via a safe `Seq.toArray` iteration.

`IPCStore<'T>` lives in its own file (`IPCStore.fs`) so unit tests import it without touching any Electron remoting code.

Each channel gets one `IPCStore` instance. IPC handlers register **eagerly at module initialization** via `do Remoting.init |> Remoting.buildHandler { ... }`. This prevents lazy-registration; any Main→Renderer event is captured as soon as the renderer bridge module loads.

Stable module-level `subscribe`/`getSnapshot` references are hoisted so `React.useSyncExternalStore` never sees new function identities across re-renders. One `[<Hook>]` per channel returns `'T voption`.

#### Consumer migration: Hooks for React state, subscriptions for Elmish dispatch

Direct React consumers (`Navbar`, `AuthStateCtx`, `FileStateCtxProvider`) read IPC snapshots via hooks. `ValueNone` (no event yet) falls back to the value from the initial request-response fetch; `ValueSome` overlays it with the latest broadcast. This keeps the initial load path intact while ending the need to register cleanup subscriptions in effects.

`App.fs` and `GitProgressSubscription.fs` (Elmish owners) keep callback subscriptions intentionally. `useSyncExternalStore` returns the current snapshot immediately on subscription; an effect-based Elmish bridge would replay a stale `ValueSome` on mount and dispatch spurious messages. Event-semantic helpers `subscribePathChange` and `subscribeGitProgressUpdate` remain public for exactly this reason. `subscribeRecentArcsUpdate`, `subscribeAuthAccountsUpdate`, and `subscribeFileTreeUpdate` are removed because their only consumers have migrated to hooks.

#### File-state context: Read-only file tree, writable selection only

`FileStateController` is narrowed: `setState` and `setFileTree` are removed. The controller exposes `setSelection` and `updateSelection` only, encoding the invariant that file-tree data is IPC-owned and not writable from the renderer.

Inside `FileStateCtxProvider`, the file tree and selection are held in separate contexts (`FileTreeCtx` and `FileSelectionCtx`). The file tree is a `Map<string, FileEntry> voption` from `useFileTreeUpdate`, converted to `FileEntry[]` in a `useMemo`. Selection is `React.useStateWithUpdater ArcSelection.empty`. The public `useFileState` hook composes both into the `FileStateController` shape, memoized on `fileTree` and `selection` identity, not on a freshly allocated combined record, so selection-only updates do not invalidate file-tree consumers.

`ArcObjectExplorerCtx` migrates from `fileStateCtx.setState` to `fileStateCtx.updateSelection`, operating on the selection slice only.

Public types `FileState` and `FileStateController` are extracted to `FileStateTypes.fs` so the API-shape test compiles without importing any Electron remoting code.

Git progress subscription is extracted from `GitWorkflow.fs` into `GitProgressSubscription.fs` so the pure workflow module (and its 41-test suite) has no dependency on the renderer bridge.


src/Electron/src/Renderer/MainUpdateRendererBridge.fs
```diff
@@ -1,53 +1,97 @@
...
+// ---------------------------------------------------------------------------
+// Event subscribe helpers for Elmish dispatch bridges.
+// These intentionally preserve "new IPC event only" semantics.
+// Do not replace them with hook snapshots unless replay-on-mount is desired.
+// ---------------------------------------------------------------------------
+
+let private makeSubscribe (store: IPCStore<'T>) (handler: 'T -> unit) : (unit -> unit) =
+    store.Subscribe(fun () ->
+        match store.GetSnapshot() with
+        | ValueSome v -> handler v
```
Freymaurer: Why do we use `ValueSome`/`ValueNone` instead of simple `Some`/`None`

src/Electron/src/Renderer/Context/FileStateContext.fs
```diff
@@ -1,65 +1,97 @@
...
-let FileStateCtx =
-    React.createContext<FileStateController> (
-        {
-            state = FileState.init ()
-            setState = ignore
-            setFileTree = ignore
-            setSelection = ignore
-        }
-    )
+module private FileStateHelper =
```
Etschbeijer: Move context and context depending functions onto their own file please

Etschbeijer (changes requested):

PR #1061, opened 2026-04-27
Remove ARCObject explorer from Electron Swate.

Etschbeijer:
Remove ARCObject explorer from Electron renderer.


src/Electron/src/Renderer/Components/MainContent/Main.fs
```diff
@@ -18,8 +17,8 @@ open Renderer.Components.MainContent.TextPreviewTarget
 open Renderer.Components.MainContent.UnknownPreviewTarget
 
 /// This can be further reduced by using the actual contexts instead of passing down the states and setters as props, but this is good enough for now
-[<ReactComponent>]
-let Main (appRootPath: ArcRootPath, pageState: PageState option, leftSidebarTarget: LeftSidebarPage) =
+[<ReactMemoComponent>]
+let Main (appRootPath: ArcRootPath, pageState: PageState option) =
```
Freymaurer, line 39: Memoizing a component that is not used inside the same file does not have any effect, as far as i know

src/Electron/src/Renderer/Context/PreviewStateContext.fs
```diff
@@ -0,0 +1,47 @@
...
+open Swate.Components.Shared
+
+type PreviewState = { PendingArcFileSave: ArcFiles option }
+with
+    static member init() : PreviewState = { PendingArcFileSave = None }
+
+type PreviewStateController = {
+    state: PreviewState
+    setState: (PreviewState -> PreviewState) -> unit
+    setPendingArcFileSave: ArcFiles option -> unit
```
Freymaurer: What is the difference between: `setPendingArcFileSave` and `setState`. As `setState` does a very similiar thing, just nested inside another record type?

Freymaurer: If this is meant to store the full arcFile with changes, i recommend - as i already said previously - to do this in Main and not Renderer.

Freymaurer (changes requested):

src/Electron/src/tests/electron.spec.ts
```diff
@@ -78,6 +78,7 @@ test.describe('Swate Electron App', () => {
     const homeButton = window.locator('[aria-label="Home"] button');
     await expect(homeButton).toBeVisible({ timeout: 30000 });
     await homeButton.click();
+    await expect(window.getByRole('button', { name: 'ARC object explorer' })).toHaveCount(0);
```
Freymaurer: Why should we test if a btn does not exist for the arc object explorer?

Freymaurer (changes requested):

caroott (approved):

src/Electron/src/Renderer/Components/MainContent/Main.fs
Etschbeijer: This should be okay, based on the docs. 

I defined it on module level and called it in another file and I did not define it as a local / nested function inside another function. Usage over files is supported, as long as the source is defined on method level.

src/Electron/src/tests/electron.spec.ts
Etschbeijer: fixed

src/Electron/src/Renderer/Context/PreviewStateContext.fs
Etschbeijer: fixed

PR #1066, opened 2026-04-28
IPC rework

caroott:
## Summary

Replace Swate's mutable renderer IPC listener store with disposable Fable.Electron receiver registrations owned by React effects and Elmish subscriptions. The old architecture used a single `IMainUpdateRendererApi` handler with fan-out dictionaries. The new design gives each one-way Main→Renderer event its own focused contract and cleanup lifecycle.

### Upgrade Fable.Electron Remoting
- **Fable.Electron** 40.6.0 → 40.9.2, **Remoting** 7.1.0 → 7.2.0, **Forge** 8.1.0 → 8.2.0
- **FSharp.Core** 10.0.102 → 10.1.203 (transitive requirement of the new bindings)
- **electron** npm 40.6.0 → 40.9.2, regenerated `package-lock.json`

### Remove `IpcMainEvent` from shared two-way contracts
- Every `IpcMainEvent` parameter stripped from `IArcVaultsApi`, `IGitApi`, `IGitLabApi`, and `IGitLfsApi`
- Three handler factories now accept `IpcMainInvokeEvent` once at registration time: `IArcVaultsApi.api`, `IGitApi.api`, `GitLfs.registerGitLfsIpc`
- Main registration uses `Remoting.fromIpcMainEvent` for window-context APIs and `Remoting.fromValue` for window-agnostic APIs (`IGitLabApi`, `IAuthApi`)
- All 40+ renderer call sites drop the `(unbox null)`

### Split one-way Main→Renderer contracts
- Monolithic `IMainUpdateRendererApi` replaced by `MainToRendererIpc` module with five focused records: `IPathChangeRendererApi`, `IRecentArcsRendererApi`, `IAuthAccountsRendererApi`, `IFileTreeRendererApi`, `IGitProgressRendererApi`
- Preload bridges registered individually per contract
- Main senders use the corresponding focused `buildProxySender` — `ArcVault.fs` sends `IPathChangeRendererApi`/`IFileTreeRendererApi`/`IRecentArcsRendererApi`, `IGitApi.fs` sends `IGitProgressRendererApi`, `IAuthApi.fs` sends `IAuthAccountsRendererApi`

### Generic disposable receiver helpers
- New `Renderer.IpcReceiver` module with `subscribeProxyReceiver<'T>` (returns a `unit → unit` disposer) and `useProxyReceiver<'T>` (Feliz hook wiring it into React effects)

### Replace `MainUpdateRendererBridge` mutable store
- Five mutable dictionaries and the singleton `Remoting.init |> buildHandler` gate replaced by direct `subscribeProxyReceiver` calls 
- `Navbar.fs` switched from `useLayoutEffectOnce` to `useEffectOnce` with proper disposable cleanup

### Close-window receiver becomes disposable
- `CloseWindowController.fs` uses `useProxyReceiver<IMainSaveBeforeQuitApi>` instead of `Remoting.init |> buildHandler`

### Centralized dialog parenting
- `pickDirectory` and `pickAbsolutePaths` now pass a parent `BrowserWindow` (via factory-captured `event`) to `dialog.showOpenDialog`, making dialogs modal to the calling window

### Remove shared `IGitLfsApi`
- `IGitLfsApi` moved out of `IPCTypes.fs` into a Main-internal helper with direct `runChannel`/`cancelChannel` functions

### Narrow `AppStateContext` to `ArcRootPath`
- Simplified context to expose only the path, reducing coupling in consumers

### Regression tests
- **`IpcReceiverDisposable.test.fs`**  proves listener init, event delivery, idempotent dispose (double-dispose fires remover once), and no event delivery after dispose; plus React mount/re-mount/unmount lifecycle with dependency change re-registration (uses polling `waitForEffect` instead of blind `Promise.sleep`)
- **`FileStateContextReload.test.fs`**  verifies Main-synced file state updates propagate correctly
- **`MainSyncedState.test.fs`**  verifies end-to-end state snapshot sync from Main to Renderer
- **`IpcArchitectureReview.test.fs`**  architectural assertion that no `Remoting.init` slips back in


src/Electron/src/Main/IPC/GitLfs.fs
Freymaurer, line 1: What does this file do? Why did you remove the `IGitLfsApi`?

src/Electron/src/Renderer/App.fs
```diff
@@ -12,38 +12,29 @@ open Swate.Components.ErrorModal
...
-let private createGetOpenPathCmd () : Cmd<Msg> =
-    Cmd.OfPromise.either
-        (fun () -> Api.ipcArcVaultApi.getOpenPath (unbox null))
-        ()
-        SetArcRootPath
-        (fun _ -> SetArcRootPath None)
-
 let private init () : Model * Cmd<Msg> =
-    Model.Empty, createGetOpenPathCmd ()
+    Model.Empty, Cmd.none
```
Freymaurer, line 45: Is this correct? Is this instead handled via subscription?

src/Electron/src/Renderer/App.fs
```diff
@@ -122,17 +106,18 @@ let private LeftActionButtons (leftSidebarTarget: LeftSidebarPage, setLeftSideba
...
-    let appCtx: StateContext<ArcRootPath> =
-        React.useMemo (
-            (fun _ -> {
-                state = model.AppState
-                setState = setAppState
-            }),
-            [| box model.AppState |]
-        )
+    let appState =
+        Renderer.MainSyncedState.useMainSyncedState {
```
Freymaurer: you can do a subscription with `React.useElmish`. This allows you to track everything via elmish state and not having elmish state and subscription and having to sync them

src/Electron/src/Renderer/App.fs
```diff
@@ -122,17 +106,18 @@ let private LeftActionButtons (leftSidebarTarget: LeftSidebarPage, setLeftSideba
...
+            initial = None
+            load = fun () -> Api.ipcArcVaultApi.getOpenPath ()
+            subscribe = Renderer.MainUpdateRendererBridge.subscribePathChange
+            onError = ignore
+            dependencies = [||]
+        }
+
+    React.useEffect ((fun () -> dispatch (ArcRootPathChanged appState.state)), [| box appState.state |])
```
Freymaurer: yeah make subscription inside elmish logic to avopid this pattern

src/Electron/src/Renderer/GitApiClient.fs
```diff
@@ -24,41 +24,41 @@ let private callIpcWith
     }
 
 let checkGitVersions () =
```
Freymaurer, line 26: personally i do not like having all this boilerplate but i guess keep it as was there before

src/Electron/src/Renderer/MainUpdateRendererBridge.fs
Freymaurer, line 1: Let us please kill this file 🔥

src/Electron/src/Renderer/MainSyncedState.fs
```diff
@@ -0,0 +1,89 @@
...
+    refresh: unit -> unit
+}
+
+[<Hook>]
+let useMainSyncedState<'T> (config: MainSyncedStateConfig<'T>) : MainSyncedState<'T> =
+    let state, setState = React.useState config.initial
+    let isLoading, setIsLoading = React.useState true
+    let activeTokenRef = React.useRef 0
+    let liveUpdateVersionRef = React.useRef 0
+    let requestVersionRef = React.useRef 0
```
Freymaurer, line 32: I understand why we have one of these. I do not understand why we have the others, can you add some comments and cases when we require these? Sounds like this heavy logic should be well documented

src/Electron/src/Renderer/App.fs
caroott: init staying Cmd.none is correct now because the subscription registers the IPC listener first, then dispatches the initial getOpenPath snapshot request

src/Electron/src/Main/IPC/GitLfs.fs
caroott: This is a Main-process helper for Git LFS operations that need progress and cancellation.
It does three things:

- runChannel event request
  - Runs Main.Git.GitLfsService.run, using the IpcMainInvokeEvent to send progress back to the renderer via event.sender.send("git-lfs:progress", ...).

- Tracks cancellation
  - It stores a cancellation flag per request.RequestId in an in-memory dictionary, and passes a cancelCheck function into the Git LFS service.

- cancelChannel requestId
  - Marks that request as canceled, so the running Git LFS operation can observe it and stop.
  
IGitLfsApi was removed because it was not really a shared renderer IPC contract anymore. The renderer does not call a standalone IGitLfsApi, it calls IArcVaultsApi.runGitLfs (IPCTypes.fs line 49). Main handles that in IArcVaultsApi.fs (line 573), enforces the active ARC path, then delegates to GitLfs.runChannel.

src/Electron/src/Renderer/MainUpdateRendererBridge.fs
caroott: done

src/Electron/src/Renderer/GitApiClient.fs
caroott: You mean the version check?

src/Electron/src/Main/IPC/GitLfs.fs
Freymaurer: But this part: `event.sender.send (GitLfsProgressChannel, [| box request.RequestId; box msg |])`, could be a Main -> Renderer ipc call, right? Because currently it is not type save

src/Electron/src/Renderer/GitApiClient.fs
Freymaurer: that all of this is more or less `fun () -> gitApi.XXX()` call

src/Electron/src/Swate.Electron.Shared/IPCTypes.fs
```diff
@@ -25,93 +25,92 @@ open IPCTypesHelper
 /// Two Way Bridge: Renderer <-> Main
 type IArcVaultsApi = {
     /// Open ARC via folder dialog. Main decides: current window / new window / focus existing.
-    openARC: IpcMainEvent -> JS.Promise<Result<string, exn>>
+    openARC: unit -> JS.Promise<Result<string, exn>>
```
Etschbeijer, line 28: Why do we no longer need the IpcMainEvent?

caroott: We no longer need IpcMainEvent in shared IPC contracts because renderer-callable APIs should not expose Main process event context. Main still gets caller context where needed through Remoting.fromIpcMainEvent, as IpcMainInvokeEvent, inside Main-only factories like IGitApi.api and ArcVaultsApi.api.

src/Electron/src/Renderer/App.fs
```diff
@@ -10,57 +10,98 @@ open Swate.Components
 open Swate.Components.Layout
 open Swate.Components.ErrorModal
 open Swate.Electron.Shared
+open Swate.Electron.Shared.IPCTypes.MainToRendererIpc
 
 type private Model = {
-    AppState: ArcRootPath
+    ArcRootPath: ArcRootPath
+    ArcRootPathLiveUpdateVersion: int
```
Freymaurer, line 19: This, again, requires comments why they exist

Freymaurer (changes requested):
one minor change then lgtm

src/Electron/src/Renderer/App.fs
caroott: added

Etschbeijer (approved):
lgtm

PR #1072, opened 2026-04-29
Discard changes (git)

caroott:
Implemented discard/undo for selected Git sidebar files.

The main pieces are:

- Added GitService.discardPaths in GitService.fs, using Git reset/restore/clean behavior and handling selected rename originals.
- Added IPC/renderer plumbing through gitDiscardPaths.
- Added DiscardSelection workflow handling in GitWorkflow.fs, including busy state, refresh, and clearing the open diff after discard.
- Added a right-click context menu to changed file rows in GitSidebar.fs. If the right-clicked file is already selected, it discards the selected set; otherwise it discards that file only.

image
image


Etschbeijer (approved):
lgtm

PR #1075, opened 2026-04-30
YAML Frontmatter for Notes

caroott:
## Summary

This PR changes notes from markdown headers to typed YAML frontmatter.

Notes now store their title, creation date, and ontology tags in a reusable `NoteFrontmatter` contract under `NoteConversion`. The markdown body stays plain text, so content like `---` inside the body is preserved instead of being treated as note metadata.

It also sets the current date as default for Note creation and reduces the date accuracy to the current day (no longer hours and minutes)

## Changes

- Added YAMLicious encoders and decoders for note frontmatter, ontology annotations, and comments.
- Updated `formatMarkdown` to write YAML frontmatter instead of `# Title`, `Date Created:`, and comma-separated tag lines.
- Added `tryDecodeMarkdownFrontmatter` for shared note parsing.
- Updated Electron note search to read only YAML-frontmatter notes and skip malformed or legacy note files.
- Added shared conversion tests for serialization, decoding, missing tags, and body preservation.
- Added Electron Core tests for YAML note reading and skipping legacy notes.
- Added Electron Renderer DTO round-trip coverage for note metadata and ontology tags.


Freymaurer:
Have not looked at the code, yet. But let us assume in 1 year we want to update the frontmatter without loosing support for legacy frontmatter. Does your code support creation and checking for legacy frontmatter? 

One easy thing would be to move encoder/decoder into a module "NoteFrontmatterV1"

caroott:
> Have not looked at the code, yet. But let us assume in 1 year we want to update the frontmatter without loosing support for legacy frontmatter. Does your code support creation and checking for legacy frontmatter?
> 
> One easy thing would be to move encoder/decoder into a module "NoteFrontmatterV1"

What legacy frontmatter do you mean? Currently, there is none. If we update the frontmatter we can split it and preserve backwards compatibility. Since we don't know how we are maybe changing it in the future we can't reliably check it at this time.

tests/Shared/Shared.Tests.fs
```diff
@@ -14,5 +14,6 @@ let example_tests = testList "example" [
 let shared = testList "Shared" [
     example_tests
     LandingTests.tests
+    NotesConversionTests.tests
```
Freymaurer: This is the wrong project for this test? Shared is run on browser and dotnet Environment. But notes reading is done in nodejs in electron. Never touching ditnet server side?

Freymaurer (changes requested):
Move tests then it is a really good PR and lgtm

src/Components/src/Notes/Editor/Conversion.fs
```diff
@@ -9,6 +9,115 @@ open Swate.Components.Landing
...
     let private notesRootFolder = "notes"
+    let private frontmatterDelimiter = "---"
+
+    type NoteFrontmatter = {
+        Title: string
+        Date: DateTime
+        Tags: ResizeArray<OntologyAnnotation> option
+    }
+
+    let private normalizeNewlines (content: string) =
```
Etschbeijer, line 20: I think this is so generic that it should be moved to a more generic helper module

PR #1076, opened 2026-04-30
Epic/feature/fix file tree selection pr minimized

Etschbeijer:
-Enable minimizing of filetree folders when an element is selected in the hierarchy
- Git LFS buttons are available
- Collape button is furthers right
- Horizontal scrolling is enabled when space of the row is not big enough
- Pill colours of Git lfs has been updated
- Transport of Git lfs metadata to renderer has been implemented
- Curspor pointer only when hovering over collapse button
- Only collapse button controlls expansion and reduction of folder children
- Closes Issue #1081


Etschbeijer:
grafik

Etschbeijer:
grafik

Gitarackur (approved):

src/Components/src/FileExplorer/FileExplorer.fs
```diff
@@ -168,42 +151,73 @@ type FileExplorer =
             else
                 Html.none
 
-        let renderLfsControls (item: FileItem) =
-            if item.IsLFS = Some true then
-                Html.div [
-                    prop.className "swt:flex swt:gap-2 swt:items-center"
+        let renderLfsStatusPill (item: FileItem) =
```
Freymaurer: Move this outside the component and make it their own private static member

Freymaurer: Remember to update naming to LFSStatusPill

src/Components/src/FileExplorer/FileExplorer.stories.tsx
```diff
@@ -74,15 +74,211 @@ export const Default: Story = {
...
+        expect(breadcrumbs).toBeTruthy();
+        expect(within(breadcrumbs!).getByText(label)).toBeTruthy();
+      });
+    };
 
     const container = await canvas.findByTestId("file-explorer-container");
+    const scrollContainer = await canvas.findByTestId("file-explorer-scroll-container");
     expect(container).toBeTruthy();
+    expect(scrollContainer).toBeTruthy();
+    expect(scrollContainer.className).toContain("swt:overflow-x-auto");
```
Freymaurer: Is it useful to check for an existing class? This still can be faulty. Better to provoke a scroll with large mock data and check if it is actually scrollable

src/Components/src/FileExplorer/FileExplorer.stories.tsx
```diff
@@ -74,15 +74,211 @@ export const Default: Story = {
...
+    await userEvent.click(resume);
+    await expectBreadcrumbContains("resume.pdf");
 
-    const folder = await canvas.findByText("My Files");
-    await userEvent.click(folder);
+    const folderNameButton = await canvas.findByRole("button", { name: "My Files" });
+    let expandFolderButton = await canvas.findByRole("button", { name: "Expand My Files" });
+    expect(getComputedStyle(folderNameButton).cursor).not.toBe("pointer");
+    expect(getComputedStyle(expandFolderButton).cursor).toBe("pointer");
+    expect(expandFolderButton.className).toContain("swt:h-5");
```
Freymaurer: Yeah i think i do not really like checking for styles. These can easily be ovewritten by whatever other styling.

src/Components/src/FileExplorer/FileExplorer.stories.tsx
Freymaurer, line 1: I think i really do not like these new test cases. They are extremely specific. Testing for specific colors and icons in multiple places. If we ever update this component we will have to update a lot of arguments here.

For some it might make sense to check for the reason the styles are applied. E.g. is the container actually scrollable for large mock data. For some like button color i am not sure if this level of testing is useful. Maybe better to integrate screenshot tests if we want to check.

@caroott @Etschbeijer @Gitarackur .

src/Electron/src/Main/FileTreeCreator.fs
Freymaurer, line 1: You did not use `git lfs ls-files -j` for git lfs information access. In my opinion this is the safest and most performant way to get full information about git lfs files.

Please get in contact with @caroott about this issue and cross check with her logic in Git/*lfs*.fs files to avoid code duplication. In addition, add any required git logic into those files and use them here. Do not use this file as good example for code structure, it should never have been root level and some functions should be moved into io helpers. 

But do not refactor in this PR, just do not make it worse.

src/Electron/src/Swate.Electron.Shared/FileIOTypes.fs
```diff
@@ -7,6 +7,9 @@ type FileEntry = {
     isDirectory: bool
     path: string
     isLfs: bool option
+    isLfsPointer: bool option
```
Freymaurer: Create a record type mimicking the git lfs ls-files -j json information and use it as option for each file. If the option.IsSome it must be lfs

Freymaurer (changes requested):

src/Components/src/FileExplorer/FileExplorer.stories.tsx
Gitarackur: I do think it is useful because, in testing, it is more viable to test how users interact with interfaces. The focus should be on the user's behavior with the interface at the end of rendering. So it's more of simulating a testing component behavior when a user interacts with the component (the user wouldn't know the scroll height of the filetree on the file component, but they would see a scroll container or see a style change).

So, if the behaviour of the component changes (if we change it), then the test should fail, and through that, we update the test to its new behaviour.

src/Components/src/FileExplorer/FileExplorer.stories.tsx
Etschbeijer: I am with you, I will remove most of them but keep the scroll tests and so, that are required for usability.

src/Components/src/FileExplorer/FileExplorer.fs
Etschbeijer: fixed

src/Components/src/FileExplorer/FileExplorer.stories.tsx
Etschbeijer: removed them

src/Electron/src/Swate.Electron.Shared/FileIOTypes.fs
Etschbeijer: will be solved with the issue #1083

src/Electron/src/Main/FileTreeCreator.fs
Etschbeijer: Will be solved in #1083

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
```diff
@@ -0,0 +1,98 @@
+module Swate.Components.FileExplorerContextMenu
```
Freymaurer: Wrong namespace

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
```diff
@@ -0,0 +1,98 @@
...
+
+open Fable.Core
+open Fable.Core.JsInterop
+open Feliz
+open Swate.Components
+open Swate.Components.FileExplorerTypes
+
+let private copyPathToClipboard (path: string) =
```
Freymaurer: Private module

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
```diff
@@ -0,0 +1,98 @@
...
+        onItemClick: (FileItem -> unit) option,
+        onContextMenu: (FileItem -> Swate.Components.FileExplorerTypes.ContextMenuItem list) option,
+        dispatch: FileExplorerLogic.Msg -> unit
+    ) =
+    let customItems =
+        onContextMenu |> Option.map (fun fn -> fn item) |> Option.defaultValue []
+
+    defaultContextMenuItems (item, model, onItemClick, dispatch) @ customItems
+
+let toComponentMenuItem (item: Swate.Components.FileExplorerTypes.ContextMenuItem) =
```
Freymaurer: This is a reactcomponent. See design docs. Wrong name. Wrong usage. Missi g attribute.

src/Components/src/FileExplorer/FileExplorerItem.fs
```diff
@@ -0,0 +1,285 @@
+namespace Swate.Components
```
Freymaurer: Wrong namespace

src/Components/src/FileExplorer/FileExplorerItemHelper.fs
```diff
@@ -0,0 +1,15 @@
+module Swate.Components.FileExplorerItemHelper
```
Freymaurer: Wron module name. Rename to Swate.Components.FileExplorer.Helper

Freymaurer (changes requested):

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
Etschbeijer: fixed

src/Components/src/FileExplorer/FileExplorerItem.fs
Etschbeijer: fixed

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
Etschbeijer: It is not a reactComponent

src/Electron/src/Main/FileTreeCreator.fs
caroott: I would solve it in this PR tbh. There is no reason to introduce the logic here at the wrong place just to correct it later. This isn't a major addition to the existing lfs implementation.

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
Etschbeijer: fixed

src/Components/src/FileExplorer/FileExplorerItemHelper.fs
Etschbeijer: fixed

src/Components/src/FileExplorer/FileExplorer.stories.tsx
Freymaurer: But testing for an existing class is the exact oposite of testing what the user experiences. Just because the class exists, does not mean, the expected UX is shown.

- Testing for a class (just code testing)
- Testing if a container is actually scrollable with large data (What the user will experience)

So we both agree that we should test, whatever the user sees and interacts with.

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
Freymaurer: ahhhhh! This `Swate.Components.ContextMenuItem` suggested to me, that it is in fact a `ReactElement`. But it creates a class? Record type? Sorry for the confusion, i was mislead by the pre-design docs namespace. My fault.

src/Components/src/FileExplorer/FileExplorerContextMenu.fs
Freymaurer: Okay, so now that i understand the scope of this file. It should be: `module Swate.Components.FileExplorer.Helper` or `module Swate.Components.FileExplorer.ContextMenuHelper`

- It MUST NOT contain a `ReactElement`.
- IT MAY contain as many helper functions, ordered in whatever scope you want, as long as you keep a single root level module (the one on top of the file.). You MAY introduce further modules, as long as not all functions are inside a single one.

Freymaurer: ignore my other comments for this file, the were made under wrong assumptions. if you ever find a comment from me misleading or with an error, please contact me, so we can see which other comments must be updated.

src/Components/src/FileExplorer/FileExplorerItem.fs
Freymaurer: no? it should be `Swate.Components.FileExplorer`, right? Correct me if i am wrong!

src/Components/src/FileExplorer/FileExplorerItemHelper.fs
Freymaurer: no? Do you maybe have local changes?

Freymaurer: Will stop review at this point, until clarified.

Freymaurer (changes requested):

src/Components/src/FileExplorer/FileExplorer.stories.tsx
Gitarackur: The existing class is synonymous with an experience, just like how styling a div box with a red colour makes it red, and the user sees the red colour. same with the overflow class (shows an overflow scroll bar)

If these don't work as intended, then it tells about a much bigger problem (class mismatch, class hierarchy, or config-based issues (Tailwind, etc.) or even a wrong use of class). In that case, it should be solved and resolved (because the problem wouldn't exist if everything worked fine).

Failing tests shouldn't be feared in any sense. It's a vital indicator and a watchdog in test-driven development

src/Electron/src/Renderer/Components/FileExplorerLfs.fs
```diff
@@ -0,0 +1,72 @@
...
+    let size = sizeBytes |> Option.map int64
+    size, (size |> Option.map FileTree.formatSize)
+
+let getFileTreeNodeLfsState (node: FileTreeNode) : FileTreeNodeLfsState =
+    let lfsSize, lfsSizeFormatted = mapLfsSize (node.lfs |> Option.map _.size)
+
+    {
+        IsLFS = node.lfs |> Option.map (fun _ -> true)
+        IsLFSPointer = node.lfs |> Option.map (fun info -> not info.checkout)
+        Downloaded = node.lfs |> Option.map _.checkout
```
caroott: That looks wrong. `git lfs ls-files -j` has a separate downloaded field, and checkout means checked out in the working tree. A fetched-but-still-pointer file would be shown as “not downloaded” even though the LFS object exists locally. This should probably be `Downloaded = node.lfs |> Option.map _.downloaded`

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -22,6 +25,64 @@ type OutboundPushPlan =
...
+    checkout: bool option
+    downloaded: bool option
+    ``oid_type``: string option
+    oid: string option
+    version: string option
+}
+
+type private GitLfsLsFilesResponseDto = { files: GitLfsLsFileInfoDto[] option }
+
+let private tryDecodeGitLfsLsFileInfo (entry: GitLfsLsFileInfoDto) : GitLfsLsFileInfo option =
```
caroott: We already have a JSON parser dependency right? Why not use that here?

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -22,6 +25,64 @@ type OutboundPushPlan =
 let private maxLfsPointerProbeBytes = 1024L
 let mutable private cachedSystemInstalled = false
 
+[<Literal>]
+let private lfsLsFilesTimeoutMs = 15000
+
+type private GitLfsLsFileInfoDto = {
```
caroott: Also, using a json parser would probably get rid of this dto

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -138,6 +199,66 @@ let isSystemInstalled () : JS.Promise<bool> =
...
+                if String.IsNullOrWhiteSpace stdoutText then
+                    return Dictionary<string, GitLfsLsFileInfo>()
+                else
+                    return parseLsFilesByRelativePath stdoutText
+        with _ ->
+            return Dictionary<string, GitLfsLsFileInfo>()
+    }
+
+/// Enriches a single file entry with Git LFS metadata from `git lfs ls-files -j`.
+let withFileEntryLfsMetadata
```
caroott: You moved too much there i think. Those two functions still belong in the FileTreeCreator or some helper there.
`parseLsFilesByRelativePath` mostly does belong in GitLfsService, because it is parsing the output contract of a Git LFS command. That is Git/LFS-specific logic, not file explorer logic.
Those two functions know about `FileEntry` and absolute-vs-repo-relative file tree enrichment, which is not Git LFS service anymore.

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -22,6 +25,64 @@ type OutboundPushPlan =
...
+            checkout = checkout
+            downloaded = downloaded
+            ``oid_type`` = oidType
+            oid = oid
+            version = version
+        }
+    | _ ->
+        None
+
+let private parseLsFilesByRelativePath (stdoutText: string) : Dictionary<string, GitLfsLsFileInfo> =
```
caroott: I would separate parsing from indexing here and compose the info in `tryGetLsFilesByRelativePath`

Something like this:
```
let private parseLsFiles stdoutText : GitLfsLsFileInfo[] = ...

let private indexByRelativePath files =
    let map = Dictionary<string, GitLfsLsFileInfo>()
    files |> Array.iter (fun info -> map.[normalizeSeparators info.name] <- info)
    map
```

caroott (changes requested):

src/Electron/src/Main/Git/GitLfsService.fs
Freymaurer: This function is called "...ByRelativePath" but does not require any relative path? Please rename

Freymaurer (changes requested):

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed

src/Electron/src/Renderer/Components/FileExplorerLfs.fs
Etschbeijer: Fixed in commit 3c25cc9e6111221ae33c07281aa466c395e38452

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: Fixed in commit 3c25cc9e6111221ae33c07281aa466c395e38452

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: Fixed in commit 3c25cc9e6111221ae33c07281aa466c395e38452

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed

Etschbeijer: Fixed in commit 3c25cc9e6111221ae33c07281aa466c395e38452

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: Fixed in commit 3c25cc9e6111221ae33c07281aa466c395e38452

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -29,58 +28,50 @@ let mutable private cachedSystemInstalled = false
...
-            size = size
-            checkout = checkout
-            downloaded = downloaded
-            ``oid_type`` = oidType
-            oid = oid
-            version = version
-        }
-    | _ ->
-        None
+let private gitLfsLsFileInfoDecoder : Decoder<GitLfsLsFileInfo option> =
```
caroott: move json decoding to appropriate place later

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -29,58 +28,50 @@ let mutable private cachedSystemInstalled = false
...
+        let name = get.Optional.Field "name" Decode.string
+        let size = get.Optional.Field "size" Decode.float
+        let checkout = get.Optional.Field "checkout" Decode.bool
+        let downloaded = get.Optional.Field "downloaded" Decode.bool
+        let oidType = get.Optional.Field "oid_type" Decode.string
+        let oid = get.Optional.Field "oid" Decode.string
+        let version = get.Optional.Field "version" Decode.string
+
+        match name, size, checkout, downloaded, oidType, oid, version with
+        | Some name, Some size, Some checkout, Some downloaded, Some oidType, Some oid, Some version ->
```
caroott: you accept it only if all fields are present. there is no true optional handling for missing fields.

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -29,58 +28,50 @@ let mutable private cachedSystemInstalled = false
...
-        |> Array.iter (fun fileEntry ->
-            match tryDecodeGitLfsLsFileInfo fileEntry with
-            | Some info when not (String.IsNullOrWhiteSpace info.name) ->
-                let relativePath = PathHelpers.normalizeSeparators info.name
-                filesByRelativePath.[relativePath] <- { info with name = relativePath }
-            | _ -> ()
-        )
-    with _ ->
-        ()
+let private indexByRelativePath (files: GitLfsLsFileInfo[]) : Dictionary<string, GitLfsLsFileInfo> =
```
caroott: indexUsing...?
idexBy sounds like they are the input

caroott (changes requested):

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed in commit 429496fef3c32ba89840b5066369da70d72403a9

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed in commit 558ecb66ecd5cb569682bbfe6471ab516e5f5593

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed with commit b4da86c1bc0aebcb46c1ff69a870bc26604a76db

Issue #1073, opened 2026-04-29
[Dev] improve `src/Components/src` components file placement


Freymaurer:
Development in our components project is amazingly fast and it is increasingly difficult to keep a structured overview of what we have. I propose sorting the components currently under `src/Components/src` into folders, describing their implementation level. 

> [!IMPORTANT]
> This does NOT change the component folder itself, as i think this is currently well designed via the design doc.

--- 

### Primitive Components

Low level components, that are meant to be reused by others. 
- BaseModal
- Actionbar
- CardGrid
- ComboBox

### Composite Components

Higher level components, but still generic and usable for multiple higher level components

- Table
- MarkdownText?
- Authentication
- Layout

### ARC Components

Components specifically designed for usage on ARC types

- TermSearch
- AnnotationTable
- TemplateBrowser

### Page Components

Final Screen components, consisting of multiple complex components. Meant to be dropped directly into app with only some wiring missing.

- ArcFileEditor
- GitSidebar

---

Requesting feedback @caroott @Etschbeijer

caroott:
Sounds good, but i wouldn't say implementation level, that sound like they are not fully implemented yet. Category would be better i think.
We also need to keep in mind that they could potentially switch categories. Authentication for example may be a Composite component for now but could very fast become an ARC Component.

Gitarackur:
I would suggest we should not sort mainly by implementation level but sort by domain ownership first (Domain-Driven Design) and use the implementation level only within the shared UI.

- Primitive Components
- Composite Components
- ARC Components
- Page Components

I would suggest:
```
- Shared/
  - Primitives/
  - Composites/
  - Providers/
- ARC/
  - Annotation/
  - Templates/
  - Files/
  - DataMaps/
- Identity/
- Workbench/
- Util/
```

```
- BaseModal, Actionbar, CardGrid, ComboBox → Shared/Primitives
- Table, MarkdownText, Layout → Shared/Composites
- Authentication → Identity/Authentication
- TermSearch, AnnotationTable → ARC/Annotation
- TemplateBrowser, Template → ARC/Templates
- ARCSelector, FileExplorer, ArcFileEditor → ARC/Files
- DataMapTable → ARC/DataMaps
- GitSidebar, QuickAccessButton → Workbench
- ThemeProvider → Shared/Providers
```

With domain-driven design, it'll be easy to find and classify updates as well as add new features (to their respective domains) without issues

Freymaurer:
@caroott 

[...]

PR #1089, opened 2026-05-06
Refactor code locations, make layout components generic, create ThemeSelector, create SettingsPage

Freymaurer:
- Increase robustness of TermSearchQueryProvider
- Move `Components/**/Metadata/Generic.fs` with generic layout helper components to: `src\Components\src\GenericComponents\LayoutComponents.fs`
  - I have used the namespace `Swate.Components`, instead of design docs `Swate.Components.GenericComponents` in respect of possible changes due to #1073. As i thought it might be a good idea to keep all "primitives" in root namespace? @caroott @Etschbeijer 
- Rename `ThemeProvider` namespace to `Theme` namespace and add `ThemeSelector` Component (implementation taken from Client/** and cleaned)
- add PageComponents/** namespace and add `SettingsPage` component taking minimal logic from Client/** and cleaned.
- Add storybook examples in respect to #1073 in new namespaces: "'Page Components/SettingsPage'" and "'Composite Components/ThemeSelector'"
- Fix sidebar toggle icon to correctly display left and right
- Increase lazy component robustness by calling explicit argument names, to provoke compile time errors on argument rename

### DIFF
--- a/src/Components/src/Layout/Layout.fs
+++ b/src/Components/src/Layout/Layout.fs
@@ -108,7 +108,7 @@ type Layout =
         ]
 
     [<ReactComponent>]
-    static member LeftSidebarToggleBtn(?activeBorderStyle: bool) =
+    static member LeftSidebarToggleBtn() =
         let ctx = useLeftSidebarCtx ()
 
         Layout.LayoutBtn(
@@ -123,7 +123,7 @@ type Layout =
         )
 
     [<ReactComponent>]
-    static member RightSidebarToggleBtn(?activeBorderStyle: bool) =
+    static member RightSidebarToggleBtn() =
         let ctx = useRightSidebarCtx ()
 
         Layout.LayoutBtn(
--- a/src/Components/src/Layout/Layout.fs
+++ b/src/Components/src/Layout/Layout.fs
@@ -110,14 +110,13 @@ type Layout =
     [<ReactComponent>]
     static member LeftSidebarToggleBtn(?activeBorderStyle: bool) =
         let ctx = useLeftSidebarCtx ()
-        let showIsActive = activeBorderStyle |> Option.map (fun _ -> ctx.state)
 
         Layout.LayoutBtn(
             iconClassName =
                 (if ctx.state then
-                     "swt:fluent--panel-right-48-filled"
+                     "swt:fluent--panel-left-48-filled"
                  else
-                     "swt:fluent--panel-right-48-regular"),
+                     "swt:fluent--panel-left-48-regular"),
             tooltip = "Toggle left sidebar",
             tooltipClassName = "swt:tooltip-left",
             onClick = (fun () -> ctx.setState (not ctx.state))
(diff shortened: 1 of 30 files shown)


src/Components/src/Layout/Layout.fs
```diff
@@ -110,14 +110,13 @@ type Layout =
     [<ReactComponent>]
     static member LeftSidebarToggleBtn(?activeBorderStyle: bool) =
         let ctx = useLeftSidebarCtx ()
-        let showIsActive = activeBorderStyle |> Option.map (fun _ -> ctx.state)
```
caroott, line 113: `activeBorderStyle` is still an optional parameter but never used? Same for `RightSidebarToggleBtn`.

src/Components/src/TermSearch/TermSearchConfigProvider.fs
```diff
@@ -10,12 +11,89 @@ open Swate.Components.TermSearch.TermSearchActiveKeysContext
...
+    [<Emit("Array.isArray($0)")>]
+    let isJsArray (value: obj) : bool = jsNative
+
     [<Literal>]
     let TIB_PREFIX = "TIB_"
 
     [<Literal>]
     let TIB_DATAPLANT_COLLECTION_KEY = "DataPLANT"
 
+    let sanitizeStringArray (arr: string[]) =
```
caroott: Do you have tests for the sanitization logic somewhere?

src/Components/src/TermSearch/TermSearchConfigSetter.fs
```diff
@@ -43,8 +44,17 @@ type TermSearchConfigSetter =
         let activeKeysCtx = useTermSearchActiveKeysCtx ()
         let allKeysCtx = useTermSearchAllKeysCtx ()
 
+        let activeKeysState =
+            if isNullOrUndefined (box activeKeysCtx.state) then
```
caroott, line 49: Is this not already checked here?
```fsharp
let sanitizeActiveKeysState
    (fallbackState: TermSearchConfigLocalStorageActiveKeysContext)
    (state: TermSearchConfigLocalStorageActiveKeysContext)
    : TermSearchConfigLocalStorageActiveKeysContext * bool =
    if isNullOrUndefined (box state) then
        fallbackState, true
    ...
```

src/Components/src/TermSearch/TermSearchConfigSetter.fs
```diff
@@ -43,8 +44,17 @@ type TermSearchConfigSetter =
...
+                TermSearchConfigLocalStorageActiveKeysContext.init ()
+            else
+                activeKeysCtx.state
+
+        let activeKeys =
+            activeKeysState.activeKeys |> Option.ofObj |> Option.defaultValue [||]
```
caroott, line 55: Same here. Are those 3 checks for basically the same?

src/Components/src/TermSearch/TermSearchConfigProvider.fs
Freymaurer: to be honest, i was stressed and had not much time left to fix an issue. so i gave AI the task and just trusted when the results where fine. Good thing pointing this out, will rework and actually tackle this task myself.

src/Components/src/Layout/Layout.fs
```diff
@@ -108,32 +108,30 @@ type Layout =
...
                  else
-                     "swt:fluent--panel-right-48-regular"),
+                     "swt:fluent--panel-left-48-regular"),
             tooltip = "Toggle left sidebar",
             tooltipClassName = "swt:tooltip-left",
             onClick = (fun () -> ctx.setState (not ctx.state))
         )
 
     [<ReactComponent>]
-    static member RightSidebarToggleBtn(?activeBorderStyle: bool) =
```
caroott, line 127: Layout.stories still has the leftover activeBorderStyle i think

Issue #1086, opened 2026-05-05
Add delete functionality to the Swate Electron file explorer context menu


Etschbeijer:
## Summary

Add a delete operation to the Swate Electron file explorer context menu, allowing users to delete files or entries directly from the UI.

## Motivation

The file explorer currently supports adding items, but users also need a convenient way to remove files or entries without leaving Swate Electron. Delete functionality should behave consistently with the existing add operation and keep the in-memory model synchronized where needed.

### Delete operation behavior

* Add a delete action to the file explorer context menu.
* The delete operation should be performed in-memory and written directly on disk, similar to the existing add operation.
  * Existing in-memory changes MUST stay in-memory. The in-memory ARC is still "dirty".

### Scope of deletion detection

* Deletions recognized by file watcher should be propagates to the relevant components.
* Deletions to non-arc files (e.g. README.md, dataset) MUST be propagated to the filetree. This MUST be done without flickering or full reload of the file tree component.
* Deletions to ARC files must trigger ARC merge (WIP) and store the merged ARC in-memory. This MUST trigger a check inside Renderer to verify, that any open files, still exist and if not it MUST update PageState to a fallback.

### UI changes

* Delete operations SHOULD use a iconofy fluent UI "bin" icon
* Add a delete button to the file explorer UI, like the plus button of the add operation. Clicking this button MUST open a confirmation modal to verify deletion, before triggering deletion workflow.
* Delete option must be available via context menu, in red text color.

## Acceptance Criteria

* Users can delete files or entries from the file explorer context menu.
* Users can delete files or entries using a bin button with delete options.
* Deletions are written directly to disk while keepin in-memory ARC changes in memory.
* External deletions are correctly forwarded.
* The file explorer UI refreshes correctly after a successful deletion. There MUST NOT be any flickering or full reload.

## Notes

This should follow the same architectural pattern as the existing add operation wherever possible.

PR #1090, opened 2026-05-07
Filetree ContextMenu button for deletion

Etschbeijer:
- Add the basic logic for Delete button of file tree context menu
- Deletion is directly written on disk, not only in memory
- When a pure in memory existing file would be delted, it will be deleted unless its path cannot be found. Then an error is thrown.
- The flow is the following: Remove file in memory -> Write to disk -> read in again ARC -> Merge with InMemoryModel -> Rerendering


Etschbeijer:
Closes Isssue #1086

caroott:
> Closes Isssue #1090

isn't this the PR itself?

caroott:
>     * Deletion is directly written on disk, not only in memory
>     * When a pure in memory existing file would be delted, an error is thrown

Is it visible for someone who is using the ARCitect if the file they are working on is only in memory or on disk as well? I think it could be quite confusing otherwise if two files look similar but only one can be deleted.

Freymaurer:
> >     * Deletion is directly written on disk, not only in memory
> >     * When a pure in memory existing file would be delted, an error is thrown
> 
> Is it visible for someone who is using the ARCitect if the file they are working on is only in memory or on disk as well? I think it could be quite confusing otherwise if two files look similar but only one can be deleted.
> 
> 

I agree this is also bad UX. As user you want to be able to delete in memory files too. But displaying in-memory only files with a marker/color/different icon could be an awesome idea! Similiar to Git changes, showing updated and added files, so users are reminded to save. But this will be complex logic and is maybe something we need to adress by revisiting the full filetree again.

Etschbeijer:
> > Closes Isssue #1090
> 
> isn't this the PR itself?

Fixed

Etschbeijer:
> > > ```
> > > * Deletion is directly written on disk, not only in memory
> > > * When a pure in memory existing file would be delted, an error is thrown
> > > ```
> > 
> > 
> > Is it visible for someone who is using the ARCitect if the file they are working on is only in memory or on disk as well? I think it could be quite confusing otherwise if two files look similar but only one can be deleted.
> 
> I agree this is also bad UX. As user you want to be able to delete in memory files too. But displaying in-memory only files with a marker/color/different icon could be an awesome idea! Similiar to Git changes, showing updated and added files, so users are reminded to save. But this will be complex logic and is maybe something we need to adress by revisiting the full filetree again.

So, I would the user just let delte the in-memory file and keep in mind / write an issue for the second part, highlighting in-memory changed files.

Etschbeijer:
Enable in memory deletion

Freymaurer (approved):
merge now and we do the discussed refactor when we are back in epic with the component

Issue #1051, opened 2026-04-17
[Electron-Bug] Widgets


Freymaurer:
# Add Building Block

- [x] widget closes after clicking "Add Column"
- [ ] design diverges from src/Client/**

# Template

- [x] No loading view for template widget.
- [x] App crashes when importing templates (tested with multiple templates, will further investigate)
- [x] Missing refresh button
- [x] Missing stored information about when to re-request templates.

# Data Annotator

- [ ] is not correctly virtualized, all nodes are rendered
- [ ] Updating seperator creates a notification element, which pushes action buttons out of the widget visually
- [ ] Creating errors (input/output already used and "autodetect" creates error notification, which pushes action buttons out of the widget visually
- [ ] Seperator update controller should be visually grouped
- [ ] "Autodetect" is not descriptive and misses information

Freymaurer:
**Next Step**

- [x] Isolize Tempalte Widget title
- [x] Move refresh button to template list header
- [x] Make disabled state replace filter and template browser
- [x] Move new split components into their own files and clean up TemplateWidget.

Freymaurer:
> App crashes when importing templates (tested with multiple templates, will further investigate)

Was due to `t.Columns.Count`. fixed by changing it to t.ColumnCount!

This seesm to happen when you use `thisIsArcFile.ArcTables.AddTable(xxx)` and run the code below on the new table. Will investigate with @HLWeil 

```fsharp
let hasColumns =
    activeView.TryGetActiveTable(arcFileState)
    |> Option.map (fun t -> t.Columns.Count > 0)
    |> Option.defaultValue false

```

Freymaurer:
Done!

PR #1095, opened 2026-05-11
Feature/update filepicker widget

Freymaurer:
- Update File Picker widget #1051 
- Add drag and drop and remove-drop zone for widget
- Enable picking more files, while files are still picked.

image

### DIFF
--- a/src/Components/src/ArcFileEditor/ArcFileEditor.fs
+++ b/src/Components/src/ArcFileEditor/ArcFileEditor.fs
@@ -22,21 +22,44 @@ type private LazyComponents =
 
     [<ReactLazyComponent>]
     static member LazyDataMap(datamap: DataMap, setDatamap: DataMap -> unit) =
-        DataMapTable.DataMapTable(datamap, setDatamap)
+        DataMapTable.DataMapTable(datamap = datamap, setDatamap = setDatamap)
 
     [<ReactLazyComponent>]
     static member LazyBuildingBlockWidget
         (arcFile: ArcFiles, activeTableIndex: int option, setArcFile: ArcFiles -> unit)
         =
-        Widgets.BuildingBlockWidget.Main(arcFile, activeTableIndex, setArcFile)
+        Widgets.BuildingBlockWidget.Main(
+            arcFile = arcFile,
+            activeTableIndex = activeTableIndex,
+            setArcFile = setArcFile
+        )
 
     [<ReactLazyComponent>]
     static member LazyTemplateWidget(arcFile: ArcFiles, activeTableIndex: int option, setArcFile: ArcFiles -> unit) =
-        Widgets.TemplateWidget.TemplateWidget(arcFile, activeTableIndex, setArcFile)
+        Widgets.TemplateWidget.TemplateWidget(
+            arcFile = arcFile,
+            activeTableIndex = activeTableIndex,
+            setArcFile = setArcFile
+        )
 
     [<ReactLazyComponent>]
-    static member LazyArcFileMetadata(arcFile, setArcFile) =
-        Metadata.ArcFileMetadata.ArcFileMetadata(arcFile, setArcFile)
+    static member LazyFilePickerWidget
+        (
+            arcFile: ArcFiles,
+            activeTableIndex: int option,
+            setArcFile: ArcFiles -> unit,
+            onPickPaths: unit -> Fable.Core.JS.Promise<string[]>
+        ) =
+        Widgets.FilePickerWidget.Main(
+            arcFile = arcFile,
+            activeTableIndex = activeTableIndex,
+            setArcFile = setArcFile,
+            onPickPaths = onPickPaths
+        )
+
+    [<ReactLazyComponent>]
+    static member LazyArcFileMetadata(arcFile: ArcFiles, setArcFile: ArcFiles -> unit) =
+        Metadata.ArcFileMetadata.ArcFileMetadata(arcFile = arcFile, setArcFile = setArcFile)
 
 [<Erase; Mangle(false)>]
 type Main =
@@ -48,43 +71,8 @@ type Main =
             prop.text text
         ]
 
-    [<ReactComponent>]
-    static member private LazyDataMapView(datamap: DataMap, setDatamap: DataMap -> unit) =
-        React.Suspense(
-            [ LazyComponents.LazyDataMap(datamap, setDatamap) ],
-            fallback = Main.LazyFallback("Loading DataMap view...")
-        )
-
-    [<ReactComponent>]
-    static member private LazyBuildingBlockWidget
-        (arcFile: ArcFiles, activeTableIndex: int option, setArcFile: ArcFiles -> unit)
-        =
-        React.Suspense(
-            [
-                LazyComponents.LazyBuildingBlockWidget(arcFile, activeTableIndex, setArcFile)
-            ],
-            fallback = Main.LazyFallback("Loading building block widget...")
-        )
-
-    [<ReactComponent>]
-    static member private LazyTemplateWidget
-        (arcFile: ArcFiles, activeTableIndex: int option, setArcFile: ArcFiles -> unit)
-        =
-        React.Suspense(
-            [
-                LazyComponents.LazyTemplateWidget(arcFile, activeTableIndex, setArcFile)
-            ],
-            fallback = Main.LazyFallback("Loading template widget...")
-        )
-
-    [<ReactComponent>]
-    static member private LazyArcFileMetadata(arcFileState, setArcFileState) =
-        React.Suspense(
-            [
-                LazyComponents.LazyArcFileMetadata(arcFileState, setArcFileState)
-            ],
-            fallback = Main.LazyFallback("Loading metadata view...")
-        )
+    static member LazyLoaderWithMessage(lazyComponent: ReactElement, message: string) =
+        React.Suspense([ lazyComponent ], fallback = Main.LazyFallback(message))
 
     [<ReactMemoComponent(AreEqualFn.FsEqualsButFunctions)>]
     static member private AddRowsFooterView(props: AddRowsFooterViewProps) =
@@ -130,7 +118,11 @@ type Main =
         (activeView: ActiveView, arcFileState: ArcFiles, setArcFileState: ArcFiles -> unit)
         =
         match activeView with
-        | ActiveView.Metadata -> Main.LazyArcFileMetadata(arcFileState, setArcFileState)
+        | ActiveView.Metadata ->
+            Main.LazyLoaderWithMessage(
+                LazyComponents.LazyArcFileMetadata(arcFileState, setArcFileState),
+                "Loading metadata..."
+            )
         | ActiveView.Table index ->
             let tables = arcFileState.Tables()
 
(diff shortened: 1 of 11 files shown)


caroott:
- Rename "Cancel" to "Clear"
- Remove "Drop Row here to remove" zone (delete button is more intuitive)
- Add Drop Zone for new files
- Make files selectable
  - add or switch to clear selected and insert selected
- Does sorting work on full path or filename?

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,101 +4,337 @@ open ARCtrl
...
+    let sortAscending (setPaths: (string[] -> string[]) -> unit) =
+        fun () -> setPaths (fun current -> current |> Array.sortBy id)
+
+    let sortDescending (setPaths: (string[] -> string[]) -> unit) =
+        fun () -> setPaths (fun current -> current |> Array.sortByDescending id)
+
+    let movePath (setPaths: (string[] -> string[]) -> unit) =
+        fun (currentIndex: int) (newIndex: int) ->
+            setPaths (fun current ->
+                if
```
caroott: this case should never be possible in the first place. If it occurs, i would throw an error.

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,101 +4,337 @@ open ARCtrl
...
+                    current
+                else
+                    let pathToMove = current.[currentIndex]
+                    let pathsWithout = current |> Array.removeAt currentIndex
+                    pathsWithout |> Array.insertAt newIndex pathToMove
+            )
 
-    static member private WidgetContainerClass =
-        "swt:flex swt:flex-col swt:gap-2 swt:p-2 swt:min-w-80 swt:max-w-[95vw]"
+    let removePath (setPaths: (string[] -> string[]) -> unit) =
```
caroott: maybe name it removePathAt. I would expect this to have a string input and remove a path with that name.

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,101 +4,337 @@ open ARCtrl
...
+
+    let clearPaths (setPaths: (string[] -> string[]) -> unit) = fun () -> setPaths (fun _ -> [||])
+
+    let sortAscending (setPaths: (string[] -> string[]) -> unit) =
+        fun () -> setPaths (fun current -> current |> Array.sortBy id)
+
+    let sortDescending (setPaths: (string[] -> string[]) -> unit) =
+        fun () -> setPaths (fun current -> current |> Array.sortByDescending id)
+
+    let movePath (setPaths: (string[] -> string[]) -> unit) =
```
caroott: maybe name it movePathByIndex, same schema as movePathById

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,101 +4,337 @@ open ARCtrl
...
-        entries |> List.mapi (fun index (_, path) -> index + 1, path)
+    let movePathById (setPaths: (string[] -> string[]) -> unit) =
+        fun (activeId: string) (overId: string) ->
+            setPaths (fun current ->
+                let oldIndex = current |> Array.tryFindIndex ((=) activeId)
+                let newIndex = current |> Array.tryFindIndex ((=) overId)
 
-    static member private disabledState(message: string) =
+                match oldIndex, newIndex with
+                | Some oldIndex, Some newIndex when oldIndex <> newIndex ->
```
caroott, line 66: why not call movePath(ById) here? This should cover this case (moving by Index)

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,101 +4,337 @@ open ARCtrl
...
+        (selectedCells: option<CellCoordinateRange>)
+        =
+        fun () ->
+
+            match arcFile.TryGetActiveTable(activeTableIndex), selectedCells with
+            | Some(_, table), Some selection ->
+                let columnIndex = selection.xStart
+                let mutable rowIndex = selection.yStart
+
+                let cellsToInsert = [|
```
caroott: This adds the pathlist up to the index matching the rowcount after the selected cell right?
So if i have 10 rows but 15 files, 5 files are omitted? Would it make sense to create new rows for the remaining files?

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,101 +4,337 @@ open ARCtrl
...
+                        | Some cell ->
+                            let nextCell = cell.UpdateMainField path
+                            let coordinate: CellCoordinate = {| x = columnIndex; y = rowIndex |}
+                            coordinate, nextCell
+                            rowIndex <- rowIndex + 1
+                        | None -> ()
+                |]
+
+                if cellsToInsert.Length = 0 then
+                    ()
```
caroott: This returns unit in fail cases as far as i can see. It should be greyed out if canInsert is false, so this shouldn't be hit anyways. But wouldn't it be better to return a warning in unexpected cases where it may happen instead of silently doing nothing?

src/Components/src/ArcFileEditor/Widgets/Main.fs
```diff
@@ -130,10 +136,19 @@ type Main =
                             prefix = "ADD_TEMPLATE"
                             content = templateWidget
                         |}
+                        WidgetType.FilePicker,
+                        {|
+                            prefix = "FILEPICKER"
+                            content = filePickerWidget
+                        |}
```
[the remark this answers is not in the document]
caroott: Is this resolved?
widgetTypes include DataAnnotator and WidgetToggleBtns has buttons for all widget types. Does this then result in an empty widget (Widget.Base)?

src/Components/src/ArcFileEditor/ArcFileEditor.fs
caroott, line 336: Hash codes as memo keys are dangerous. Ask Lukas.

src/Components/src/ArcFileEditor/Types.fs
```diff
@@ -60,6 +60,6 @@ type ArcFileEditorHeaderProps = {
 }
 
 type ArcFileEditorWidgetServices = {
```
caroott: Is this type still used?

src/Components/src/ArcFileEditor/Widgets/Main.fs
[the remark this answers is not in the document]
Freymaurer: @caroott yes this is on purpose until the next PR. Check the relevant issue. Changes are still open :)

src/Components/src/Widgets/FilePickerWidget.fs
Freymaurer: fixed in 271482bc83809ee9ff7d8110db9d3c92ea814276 (still local, will rerequest review)

Freymaurer: Actually replaced again by dndkit.moveArray to unify logic

daba0c49fe043f381b1c2d4c93b5c0445849c417

src/Components/src/Widgets/FilePickerWidget.fs
Freymaurer: fixed in daba0c49fe043f381b1c2d4c93b5c0445849c417

src/Components/src/Widgets/FilePickerWidget.fs
Freymaurer: fixed in daba0c49fe043f381b1c2d4c93b5c0445849c417

src/Components/src/Widgets/FilePickerWidget.fs
Freymaurer: unified to dndkit utility function here: daba0c49fe043f381b1c2d4c93b5c0445849c417

src/Components/src/Widgets/FilePickerWidget.fs
Freymaurer: i think it would make sense to have a unified logic for this between inserting paths and copy-paste-behavior, this is something we should discuss in the near future and i would not make part of this PR.

But i think your request will be required at some point

src/Components/src/Widgets/FilePickerWidget.fs
Freymaurer: fixed in 55bd3a084f22ddf6fcbf8e8e4d453a632ef4934c

src/Components/src/ArcFileEditor/ArcFileEditor.fs
Freymaurer: we must assume that they work, but you may give me a suggestion for an alternative depdency based on an arcfile :D

src/Components/src/ArcFileEditor/Types.fs
Freymaurer: removed in 9393dd61e4154e452abbb467191f2541dee71576

Freymaurer:
> * Add Drop Zone for new files

Sadly not useful, as drag'n'drop into a browser does only provide the filename and not a path :/

But it gave me a new idea, i'd like to implement in another PR. FilePicker might be of very little use anyways, as we could allow drag and dropping files from the file tree into an table(!). The filetree does have the information of the paths and it already shows the ARC context inside the same app. But this is something for another PR

> * Does sorting work on full path or filename?

Path, we do not even store the filename specifically

---

New Image

image

caroott:
> > * Add Drop Zone for new files
> 
> Sadly not useful, as drag'n'drop into a browser does only provide the filename and not a path :/
> 
> But it gave me a new idea, i'd like to implement in another PR. FilePicker might be of very little use anyways, as we could allow drag and dropping files from the file tree into an table(!). The filetree does have the information of the paths and it already shows the ARC context inside the same app. But this is something for another PR
> 

That is a good idea! This should also be a problem for the image markdown plugin, which uses the browser drop zone in the component. This should be solvable by this as well.

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,126 +4,374 @@ open ARCtrl
...
+                        prop.title
+                            "Insert file paths into the currently selected table cells. Requires an active table view and selected cells."
+                        prop.onClick (fun _ -> insertPaths false)
+                    ]
+                | _ ->
+                    Html.button [
+                        prop.className "swt:btn swt:btn-neutral"
+                        prop.text "Clear Selected"
+                        prop.title "Clear only the currently selected paths from the list"
+                        prop.onClick (fun _ ->
```
caroott, line 346: Clear Selected removes paths but leaves `selectedPaths` unchanged. Deleting a selected row has the same issue. The footer can stay in “selected” mode with no visible selected rows, and Insert selected can pass an empty path array into `insertPathsIntoSelectedTableCells`, which then throws at
```
if cellsToInsert.Length = 0 then
    failwith "No valid cells to insert paths into. Please check the selected range and try again."
```
Clear `selectedPaths` whenever paths are removed/cleared, or derive selection from the current path list

src/Components/src/Widgets/FilePickerWidget.fs
```diff
@@ -4,126 +4,374 @@ open ARCtrl
...
-                                Html.td [
-                                    prop.className "swt:max-w-[28rem] swt:truncate"
-                                    prop.title path
-                                    prop.text path
+                                Html.button [
+                                    prop.className "swt:btn swt:btn-xs swt:join-item"
+                                    prop.type'.button
+                                    prop.onClick (fun e ->
+                                        e.stopPropagation ()
+                                        movePath index (index - 1)
```
caroott, line 165: If you apply this to the first item, then it's moved to the bottom of the list instead of staying where it is, right?

PR #1096, opened 2026-05-11
Remove pure in-memory addition and deletion

Etschbeijer:
- Add of the filetree no longer adds only in-memory. 
- An added target must be written directly onto the disk.
- Delete of the filetree must delete only targets on disk.

### DIFF
--- a/src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
+++ b/src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
@@ -16,20 +16,17 @@ open Types
 open Helper
 
 module private FileTreeHelper =
-    let stagePendingArcFileSave (arcFile: ArcFiles option) : JS.Promise<Result<unit, exn>> = promise {
-        let pendingSaveRequestResult =
-            match arcFile with
-            | None -> Ok None
-            | Some nextArcFile ->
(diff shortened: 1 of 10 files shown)


src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
```diff
@@ -16,20 +16,17 @@ open Types
...
-        | Error saveError -> return Error saveError
-        | Ok pendingSaveRequest ->
-            let! result = Api.ipcArcVaultApi.setPendingArcFileSave pendingSaveRequest
-            return result
+
+    let saveArcFileAndOpen (arcFile: ArcFiles) : JS.Promise<Result<FileContentDTO, exn>> = promise {
+        match FileContentDTO.fromArcFile arcFile with
+        | None -> return Error(exn "Saving this file type is not supported in Electron yet.")
+        | Some request ->
+            match! Api.ipcArcVaultApi.setArcFileInMemory request with
```
caroott: You set the file in memory first before you try to save it. If the save fails it is still changed in memory and not undone

Etschbeijer: Fixed in commit 1bdbc0d7b5c5c4341d0046219662e74792733da2

PR #1100, opened 2026-05-12
Add rename button to context menu filetree and filetree row

Etschbeijer:
# Sync in-memory ARC and file system

## syntax

- **"generic" file or folder** - refers to any file with no collection state in the app
- **arc** file or folder - refers to any file 
- **local arc** - refers to the arc stored on the disc
- **in-memory arc** - refers to the arc stored in ArcVault
- **arc merge** - refers to a function call to merge changes between a local and a in-memory arc.

### add/delete/rename generic file or folder

1. file watcher updates filetree with change

### add/delete/rename local arc file or folder

Pre 1: Investigate how renaming an assay/study/run/workflow folder affects arc integrity, as the identifier field SHOULD(?)/MUST(?) be equal to the folder name?

1. Use arc merge
2. file watcher updates filetree with change

### add arc file via filetree

Pre 2: Requires a new `ARC.Add<ArcFile>Async` function, similiar to `ARC.Rename<ArcFile>Async` and `ARC.Delete<ArcFile>Async`.

- Call `ARC.Add<ArcFile>Async` to add the new ArcFile to local AND in memory arc.
- file watcher triggers arc merge
    > [!WARNING]
    > this is some doubled logic, as the filewatcher will detect changes in an ARC file and trigger an arc merge, merging the new local file in the already existing in-memory file
- file watcher updates filetree

### rename arc file via filetree

=> Depends on `Pre 1`. 

If ARCtrl handles difference in folder and identifier name gracefully then

info: If we can use `fs` module we should use it, as it allows us to use the same logic for generic files too.

- Rename ArcFile folder using `fs` module
- file watcher triggers arc merge
- file watcher updates filetree

else

- Call `ARC.Rename<ArcFile>Async` to rename local and in memory ArcFile (this will update in-memory identifier and local ArcFile folder name)
- file watcher triggers arc merge
    > [!WARNING]
    > this is some doubled logic, as the filewatcher will detect changes in an ARC file and trigger an arc merge, merging the new local file in the already existing in-memory file
- file watcher updates filetree 

### delete arc fle via filetree

info: If we can use `fs` module we should use it, as it allows us to use the same logic for generic files too.

- Delete ArcFile folder using `fs` module
- file watcher triggers arc merge
- file watcher updates filetree

### DIFF
--- a/src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
@@ -11,6 +11,43 @@ open Main
 open Main.Bindings
 open Node.Api
 
+let private processDynamic: obj = emitJsExpr () "process"
+
+let private isWindowsPlatform () =
+    try
+        let platform = processDynamic?platform |> unbox<string>
+        platform = "win32"
+    with _ ->
+        false
+
+type private FileWatcherProfile = {
+    AwaitWriteFinish: bool
+    IgnoreInitial: bool
+    UsePolling: bool option
+    Interval: int option
+    BinaryInterval: int option
+}
+
+let private defaultWatcherProfile = {
+    AwaitWriteFinish = true
+    IgnoreInitial = true
+    UsePolling = None
+    Interval = None
+    BinaryInterval = None
+}
+
+// Polling on Windows increases event reliability for external xlsx edits at the cost of more filesystem polling.
+let private windowsWatcherProfile = {
+    AwaitWriteFinish = true
+    IgnoreInitial = true
+    UsePolling = Some true
+    Interval = Some 200
+    BinaryInterval = Some 400
+}
+
+let private activeWatcherProfile () =
+    if isWindowsPlatform () then windowsWatcherProfile else defaultWatcherProfile
+
 
 /// This function mutably sets the datamap on the correct parent based on the datamap parent info included in the file content DTO.
 /// It also ensures that the static hash is preserved to avoid unnecessary changes to the ARC when saving a datamap.
@@ -182,11 +219,20 @@ let createFileWatcher (path: string) =
             else
                 false
 
-    let watcher =
-        Chokidar.Chokidar.watch (
-            path,
-            Chokidar.WatchOptions(cwd = path, awaitWriteFinish = true, ignored = !^ignoreFn, ignoreInitial = true)
+    let watcherProfile = activeWatcherProfile ()
+
+    let watcherOptions =
+        Chokidar.WatchOptions(
+            cwd = path,
+            awaitWriteFinish = watcherProfile.AwaitWriteFinish,
+            ignored = !^ignoreFn,
+            ignoreInitial = watcherProfile.IgnoreInitial,
+            ?usePolling = watcherProfile.UsePolling,
+            ?interval = watcherProfile.Interval,
+            ?binaryInterval = watcherProfile.BinaryInterval
         )
+
+    let watcher = Chokidar.Chokidar.watch (path, watcherOptions)
 
     watcher
 
--- a/src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
@@ -11,44 +11,6 @@ open Main
 open Main.Bindings
 open Node.Api
 
-let private processDynamic: obj = emitJsExpr () "process"
-
-let private isWindowsPlatform () =
-    try
-        let platform = processDynamic?platform |> unbox<string>
-        platform = "win32"
-    with _ ->
-        false
-
-type private FileWatcherProfile = {
-    AwaitWriteFinish: bool
-    IgnoreInitial: bool
-    UsePolling: bool option
-    Interval: int option
-    BinaryInterval: int option
-}
-
-let private defaultWatcherProfile = {
-    AwaitWriteFinish = true
-    IgnoreInitial = true
-    UsePolling = None
-    Interval = None
-    BinaryInterval = None
-}
-
-// Polling on Windows increases event reliability for external xlsx edits at the cost of more filesystem polling.
-let private windowsWatcherProfile = {
-    AwaitWriteFinish = true
-    IgnoreInitial = true
-    UsePolling = Some true
-    Interval = Some 200
-    BinaryInterval = Some 400
-}
-
-let private activeWatcherProfile () =
-    if isWindowsPlatform () then windowsWatcherProfile else defaultWatcherProfile
-
-
 /// This function mutably sets the datamap on the correct parent based on the datamap parent info included in the file content DTO.
 /// It also ensures that the static hash is preserved to avoid unnecessary changes to the ARC when saving a datamap.
 let private setDataMapByParentInfo (arc: ARC) (dmpi: DatamapParentInfo) (dm: DataMap) : Result<unit, exn> =
@@ -199,7 +161,7 @@ let createWindow () = promise {
     return window
 }
 
-let createFileWatcher (path: string) =
+let createFileWatcher (path: string) (usePolling: bool option) =
 
     let ignoreFn =
         fun (path: string) ->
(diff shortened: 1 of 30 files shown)


src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -593,31 +798,83 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
...
+                                                    Arc = mergeResult.Arc
+                                                    PendingArcFileSave = mergeResult.PendingArcFileSave
+                                                })
+                                            )
+                        }
+                    )
+            with e ->
+                return Error e
+        }
+    renamePath =
```
caroott, line 876: After renamePath succeeds, the renderer only remaps tree selection and refreshes git status in FileTreeRename.fs (line 87). If an ARC file page is open, it still holds the old ArcFiles value. Saving uses that stale value in ArcFilePreviewTarget.fs (line 35), and paths are derived from the old identifier in ARCtrl.Helper.fs (line 138). This can reintroduce the old entity/path after a rename.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTreeRenameHelper.fs
```diff
@@ -0,0 +1,92 @@
...
+let normalizeRenameName (newName: string) =
+    validateRenameName newName
+
+let buildRenamedPath (sourcePath: string) (newName: string) =
+    buildRenamedSiblingPath sourcePath newName
+
+let private isRenameContextMenuTarget (relativePath: string) =
+    match ArcDeletePathRules.classifyRenameTarget relativePath with
+    | ArcDeletePathRules.RenamePathClassification.EntityFolderTarget _
+    | ArcDeletePathRules.RenamePathClassification.CanonicalEntityFileTarget _
```
caroott: Canonical file rename is misleading. The context menu allows rename on isa.assay.xlsx / datamap files but `tryBuildRenameDraft` redirects the operation to the parent entity folder at line 45. The modal still displays draft.Item.Name in FileTree.fs (line 337), so it can say it is renaming isa.assay.xlsx while actually renaming assays

src/Shared/PathHelpers.fs
```diff
@@ -242,3 +283,74 @@ module ArcDeletePathRules =
...
+                    | Some zone when PathHelpers.pathsEqual fileName ARCtrl.ArcPathHelper.DataMapFileName ->
+                        RenamePathClassification.CanonicalDataMapFileTarget(zone, identifier, normalizedRelativePath)
+                    | _ -> RenamePathClassification.GenericTarget normalizedRelativePath
+                | _ ->
+                    if PathHelpers.pathsEqual normalizedRelativePath ARCtrl.ArcPathHelper.InvestigationFileName then
+                        RenamePathClassification.InvestigationFileTarget normalizedRelativePath
+                    else
+                        RenamePathClassification.GenericTarget normalizedRelativePath
+
+    let isRenamePathAllowed (relativePath: string) =
```
caroott, line 328: Rename policy is duplicated and inconsistent. This allows GenericTarget renames, while the renderer context menu only allows entity/canonical targets in FileTreeRenameHelper.fs (line 25), and inline row rename is directory-only in FileTree.fs (line 230).

src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
```diff
@@ -11,6 +11,43 @@ open Main
...
+
+let private defaultWatcherProfile = {
+    AwaitWriteFinish = true
+    IgnoreInitial = true
+    UsePolling = None
+    Interval = None
+    BinaryInterval = None
+}
+
+// Polling on Windows increases event reliability for external xlsx edits at the cost of more filesystem polling.
```
caroott: This affects all file watching, not just rename. It may increase filesystem load, change event timing, produce duplicate events, or mask race conditions. If rename had a watcher race, I would expect a narrowly explained fix: “during rename, watcher emits X/Y, causing Z, so we handle it this way.” The current change reads more like “watcher was unreliable, enable polling globally.

tests/Electron.Core/IpcArchitectureReview.test.fs
```diff
@@ -96,16 +96,69 @@ Vitest.describe("IPC architecture review fixes", fun () ->
...
+            expectSourceContainsInOrder
+                arcVaultApiSource
+                [|
+                    "let private runArcDiskMutation"
+                    "do! vault.RefreshFileTree()"
+                    "match mergeReloadedArc reloadedArc"
+                |]
+        })
+
+    Vitest.test("Arc vault IPC contract and implementation expose renamePath", fun () ->
```
caroott, line 183: This test reads source files as text and asserts snippets such as:

- renameWithRetriesAsync sourceAbsolutePath targetAbsolutePath
- let private renameRetryStrategy
- | Some "EPERM"
- absence of do! vault.StopFileWatcher()
- absence of vault.StartFileWatcher()

That does not prove rename works. It proves the file contains a specific implementation shape. A harmless refactor can break the test, while a broken rename can still pass. This locks in code shape rather than proving behavior.

The retry logic has a similar feel. Retrying EPERM, EACCES, and EBUSY may be reasonable on Windows. But retrying ENOTEMPTY is suspicious because that often means “destination exists / target directory not empty,” which is not transient in the normal case. The code later maps ENOTEMPTY to “destination already exists,” so the retry path and error mapping are pulling in opposite directions. Maybe there was a real Windows rename race, but the PR does not prove it with a behavioral test.

What I would suggest instead:

- Remove the source-text assertions, or keep them only for very high-level architecture constraints.
- Add a temp-directory rename test that actually renames an ARC entity and verifies disk paths, file tree, and in-memory ARC state.
- Test behavior like “rename does not trigger stale watcher reload” rather than asserting StopFileWatcher is absent.

caroott (changes requested):
This is a bit much code for a rename button unless the intended scope is full ARC entity rename semantics. The reusable FileExplorer API addition looks ok. The bigger issue is the main-process rename/merge process and duplicated path rules. Smaller PRs or a better description would be preferable

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTreeRenameHelper.fs
Etschbeijer: fixed in commit bb7690c83e802b0357430514ae59714cf3349cbd

src/Shared/PathHelpers.fs
Etschbeijer: fixed in commit bb7690c83e802b0357430514ae59714cf3349cbd

src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
Etschbeijer: fixed in commit bb7690c83e802b0357430514ae59714cf3349cbd

tests/Electron.Core/IpcArchitectureReview.test.fs
Etschbeijer: fixed in commit bb7690c83e802b0357430514ae59714cf3349cbd

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: fixed in commit bb7690c83e802b0357430514ae59714cf3349cbd

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTreeRename.fs
```diff
@@ -0,0 +1,224 @@
...
+                            }
+
+                        match renameResult with
+                        | Ok() ->
+                            tryRemapSelectionPath renameDraft.SourcePath targetPath config.selectedTreePath
+                            |> Option.iter (fun remappedSelectionPath ->
+                                config.setSelection (ArcSelection.forTreePath (Some remappedSelectionPath))
+                            )
+
+                            match tryRemapActiveArcFilePath renameDraft.SourcePath targetPath config.pageState with
```
caroott, line 109: This now refreshes the open preview after rename, but the underlying rename still appears to only rename the folder on disk, not the ARC entity identifier. `openFile` resolves assays/NewAssay/isa.assay.xlsx by looking up ARC identifier NewAssay in FileIOHelper.fs (line 349). ARC.load still reports the assay identifier as OldAssay, since i don't see where the identifier is updated. So preview reload can fall back to reading the xlsx as plain text in IArcVaultsApi.fs (line 894), and the in-memory ARC can remain stale.

The new preview-refresh test does not catch that. It mocks reloadPreviewByPath and only verifies that "assays/NewAssay/isa.assay.xlsx" is passed in FileTreeRenameWorkflow.test.fs (line 244). It does not exercise renamePath -> ARC reload/merge -> openFile, where the identifier/path mismatch occurs.

caroott (changes requested):

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTreeRename.fs
Etschbeijer: Fixed in commit bf3e4a5d80c966a6af6e63750230a5ffc0a2fe0d

PR #1101, opened 2026-05-13
Feature/move save btn add ui improvements

Freymaurer:
- Increase sidebar actions icons size.
- Hide left sidebar + sidebar actions if no ARC is open.
- Move save button to Main navbar
- Sync save button with hasUnsavedArcChanges flag in ArcVaul
  - Send flag updates to renderer: by making setter private and creating custom setter, that sends updates
  - button is disabled if no changes exists

### DIFF
--- a/src/Electron/src/Main/IPC/IArcVaultsApi.fs
+++ b/src/Electron/src/Main/IPC/IArcVaultsApi.fs
@@ -41,9 +41,7 @@ module ArcPathValidation =
 
     let isWithinRootPath (rootPath: string) (candidatePath: string) =
         let normalizedRootPath =
-            pathDynamic?resolve (rootPath)
-            |> unbox<string>
-            |> normalizePathForComparison
+            pathDynamic?resolve (rootPath) |> unbox<string> |> normalizePathForComparison
 
         let normalizedCandidatePath =
             pathDynamic?resolve (candidatePath)
@@ -110,8 +108,7 @@ let private writeUtf8FileAsync (absolutePath: string) (content: string) : JS.Pro
 }
 
 let private readUtf8FileAsync (absolutePath: string) : JS.Promise<string> =
-    fsPromisesDynamic?readFile (absolutePath, "utf8")
-    |> unbox<JS.Promise<string>>
+    fsPromisesDynamic?readFile (absolutePath, "utf8") |> unbox<JS.Promise<string>>
 
 open ARCtrl.Contract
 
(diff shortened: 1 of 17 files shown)


src/Electron/src/Renderer/App.fs
```diff
@@ -193,13 +192,21 @@ let Main () =
                                             CloseWindowController.CloseWindowController.Subscription()
                                         |],
                                     navbar = Renderer.Components.Navbar.Main(),
-                                    leftSidebar = Renderer.Components.LeftSidebar.Main.Main(model.LeftSidebarTarget),
-                                    leftActions = LeftActionButtons(model.LeftSidebarTarget, setLeftSidebarTarget)
+                                    leftSidebar =
+                                        (if isInitializedArcVault then
+                                             Renderer.Components.LeftSidebar.Main.Main(model.LeftSidebarTarget)
+                                         else
+                                             Html.none),
```
caroott: You pass `Html.none` into an optional parameter, which is technically still `Some`. In Layout.fs, we just check for `leftContent.isSome` (?leftContent = leftSidebar). Wouldn't it be better to just pass `None` here?

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -553,8 +567,7 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
                                             with
                                             | Error mergeError -> return Error mergeError
                                             | Ok mergeResult ->
-                                                vault.arc <- Some mergeResult.Arc
-                                                vault.window.title <- mergeResult.Arc.Identifier
+                                                vault.SetArc mergeResult.Arc
```
caroott, line 572: This could fail in the following usage scenario:
1. User opens a clean ARC
2. In the left file explorer, user chooses Delete on an allowed ARC item
3. The delete modal calls Api.ipcArcVaultApi.deletePath deletePath in FileTreeDelete.fs (line 43)
4. Main deletes the file/folder from disk, reloads the ARC, merges the delete into the in-memory ARC, then only calls vault.SetArc mergeResult.Arc

caroott (changes requested):

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Freymaurer: I did not change any logic flow. This is only a unified replacement for the exact same call as was before?

Like SrtArc does exactly what the two calls before did?

src/Electron/src/Renderer/App.fs
Freymaurer: You are right. Thinking about it, passing none should create the cleaner transpiled code

src/Electron/src/Main/IPC/IArcVaultsApi.fs
caroott: Then it could already have failed before i guess?

Etschbeijer: What exactly should fail?

Freymaurer: Do you mean it does not propagate the delete correctly back to the renderer? If the file is still open?

caroott: Yes, that could be one way. I'm not sure of other combinations how to achieve this and how likely it is to even happen.

Freymaurer: This can easily happen. We Want to support the case of people deleting arc files directly in filesystem. The logic for this is detailed in the related issue and will be put into place together with the merge logic i have written

PR #1106, opened 2026-05-18
Data Annotator Widget

Freymaurer:
Continues work on #1051 for data annotator widget

- [x] is not correctly virtualized, all nodes are rendered
- [x] Updating seperator creates a notification element, which pushes action buttons out of the widget visually
- [x] Creating errors (input/output already used and "autodetect" creates error notification, which pushes action buttons out of the widget visually
- [x] Seperator update controller should be visually grouped
- [x] "Autodetect" is not descriptive and misses information
- [x] Added "Append" option.

---

image

image


src/Client/Pages/DataAnnotator/DataAnnotator.fs
```diff
@@ -8,537 +8,97 @@ open Messages
...
-            | {
-                  DataFile = Some _
-                  ParsedFile = Some _
-              },
-              true ->
-                DataAnnotator.Modal(model, dispatch, rmvFile, (fun _ -> setShowModal false), showModal, setShowModal)
-            | _, _ -> Html.none
-        ]
+
+    // let showModal, setShowModal = React.useState (false)
```
Etschbeijer, line 52: Why commented out and not deleted?

Etschbeijer (changes requested):

src/Client/Pages/DataAnnotator/DataAnnotator.fs
Freymaurer: There is some ARCitect specific logic inside this implementation. To my understanding it should not have been required, but to make recovery easier i kept it in. Will add a comment, explaining my thoughprocess

PR #1113, opened 2026-05-22
update: download lfs implementation

Gitarackur:


Etschbeijer:
There is no button to download the file actually...
I added it.

src/Electron/src/Renderer/Components/ARCHelper.fs
```diff
@@ -56,6 +56,23 @@ let runToggleLfsMark (relativePath: string) (markAsLfs: bool) = promise {
         | Error exn -> Error exn.Message
 }
 
+let runDownloadLfsFile (relativePath: string) = promise {
```
caroott: This bypasses the git service path and skips the path validation + authenticated remote setup.
I would prefer a dedicated main-side downloadLfsFile service/IPC endpoint.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
```diff
@@ -155,16 +169,22 @@ type FileTree =
...
 
-                    match result with
-                    | Ok loaded ->
-                        console.log ("[Renderer] Received data, processing...")
-                        Renderer.Components.ARCHelper.applyLoadedView pageStateCtx.setState loaded
-                    | Error errorMessage ->
-                        let fullErrorMessage = $"Could not open preview for '{item.Name}': {errorMessage}"
-                        console.log ($"[Renderer] Error: {fullErrorMessage}")
-                        Renderer.Components.ARCHelper.applyViewError pageStateCtx.setState fullErrorMessage
+                    if lfsDownloadRequired item then
```
caroott: this can race with the existing reload-on-file-tree-update

src/Components/src/Page/FileExplorer/FileExplorerGitLfsHelper.fs
```diff
@@ -23,70 +23,107 @@ let toggleLfsMark
         }
         |> Promise.start
 
-let freeLocalLfsCopy
+let private isLfs (item: FileItem) = item.IsLFS = Some true
+
+let private needsLfsDownload item = isLfs item && (item.Downloaded <> Some true || item.IsLFSPointer = Some true)
```
caroott: similar function exists in Helper.fs

src/Components/src/Page/FileExplorer/FileExplorer.fs
```diff
@@ -125,12 +125,14 @@ type FileExplorer =
             ?canCreateItem: FileItem -> bool,
             ?onCreateItem: FileItem -> unit,
             ?getItemActions: FileItem -> Swate.Components.Page.FileExplorer.Types.ContextMenuItem list,
+            ?getLfsPillAction: FileItem -> Swate.Components.Page.FileExplorer.Types.ContextMenuItem option,
```
caroott: this makes the reusable FileExplorer API more Git-LFS-specific and has the same signature as getItemActions. A generic row accessory/status action would fit the component intentions better

caroott (changes requested):

src/Components/src/Page/FileExplorer/FileExplorerGitLfsHelper.fs
Gitarackur: resolved

src/Components/src/Page/FileExplorer/FileExplorer.fs
Gitarackur: resolved

src/Electron/src/Renderer/Components/ARCHelper.fs
Gitarackur: resolved

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
Gitarackur: resolved

src/Components/src/Page/FileExplorer/FileExplorer.fs
caroott: You renamed it, but it is still connected to LFSStatusPill and only renders when `item.IsLFS = Some true`

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -1816,6 +1816,52 @@ let private discardPathspecsWithOriginals (arcPath: string) (status: StatusResul
...
+                            git
+
+                    match listingResult with
+                    | Error failure -> return Error failure
+                    | Ok listingJson ->
+                        match GitLfsService.tryFindListingForPath safePath listingJson with
+                        | Error message -> return errorResult (exn $"'{safePath}' {message}")
+                        | Ok listing when listing.checkout -> return Ok()
+                        | Ok _ ->
+                            let! pullResult =
```
caroott: One minor edge case could occur here. You don't check if there are local edits to the pointer before pulling the file. This is mostly irrelevant, since only pointer files are downloadable anyways. But if this was intentionally replaced by another valid pointer file, you would discard that change. We have `isPathCleanInStatus` to check for that.

caroott (changes requested):

src/Electron/src/Main/Git/GitService.fs
Gitarackur: all resolved

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -145,22 +198,52 @@ type FileExplorerItem =
...
+                prop.children badgeSegments
+            ]
+
+        let stopPillClick (e: Browser.Types.MouseEvent) =
+            e.preventDefault ()
+            e.stopPropagation ()
+
+        match statusAction with
+        | Some action ->
+            Html.button [
```
Etschbeijer, line 193: I think it would be best when you reuse one of the existing button definitions and pass in the fitting styling and action.
This button definition is too specific for Fileitem use case to be a component.

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -315,12 +407,14 @@ type FileExplorerItem =
...
             ?onDeleteItem: FileItem -> unit,
-            ?canDeleteItem: FileItem -> bool
+            ?canDeleteItem: FileItem -> bool,
+            ?statusAction: ContextMenuItem
         ) =
         let itemActions = defaultArg itemActions []
         let canDeleteItem = defaultArg canDeleteItem (fun (_: FileItem) -> false)
         let hasItemActions = itemActions |> List.isEmpty |> not
         let canDeleteFromFile = canDeleteItem item && onDeleteItem.IsSome
+        let hasStatusAction = statusAction.IsSome
```
Etschbeijer: Just call stausAction.IsSome where hasStatusAction is called.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/Helper.fs
```diff
@@ -112,6 +112,9 @@ let canDeleteItem (item: FileItem) =
     |> Option.map PathHelpers.normalizeCanonicalRelativePath
     |> Option.exists ArcEntityPathRules.isDeletePathAllowed
 
+let lfsDownloadRequired (item: FileItem) =
```
Etschbeijer: Don't pack a function call inside another function call without any additional logic, please.

Etschbeijer (changes requested):

PR #1118, opened 2026-06-03
Add delete tables button to swate-electron table view navbar

Etschbeijer:
* Add delete button to navbar of swate-electron table-view
* Move delete tables dialogue from components to shared so that it can be used in swate and swate-electron
* Deletes selected table and sets the metadatasheet active
* All changes are only in memory and not written to the disk immediately


Etschbeijer:
grafik

Etschbeijer:
grafik

Issue #1119, opened 2026-06-08
[Bug]: Closed filetree branch opens sometimes when another one is expanded


Etschbeijer:
### What happened?

When I expand a new part of the filetree, another part of it that I have closed beforehand, expands with it.

Goal: Only the targeted part of the filetree must expand.

### Host

Applies to All

### Host Details

Swate-Electron

PR #1120, opened 2026-06-08
Refactor lazy loading filetree

Etschbeijer:
Closes issue #1119


Etschbeijer:
* Seperated expand - collapse and materialize logic from one another
* Removed obsolete logic
* Filetree children elements of directories / parents are materialized when expanded
* Filetree children elements of directories / parents are not dematerialized when collapsed

PR #1124, opened 2026-06-10
Handle null cells in template import better

Etschbeijer:
Ignore null cells in import template because they can be empty -> use blank typed cells instead


Etschbeijer:
Problem:
grafik

Gitarackur (approved):

Freymaurer (changes requested):
When does this happen? What is the user story for this issue?

`let normalizeCells ` is called at many points in the code. If we ever forget calling it at a similiar position we run into the same error again. This looks like ducktaping the actual issue. Is this null issue an error in the json or the spreadsheet parsing? Depending on that we must fix this issue at the appropatie reading function.

In addition `normalizeCells` copies the table with fresh cells in an large arc this can become a performance issue - seeing how often this function is called.

Etschbeijer:
> When does this happen? What is the user story for this issue?
> 
> `let normalizeCells ` is called at many points in the code. If we ever forget calling it at a similiar position we run into the same error again. This looks like ducktaping the actual issue. Is this null issue an error in the json or the spreadsheet parsing? Depending on that we must fix this issue at the appropatie reading function.
> 
> In addition `normalizeCells` copies the table with fresh cells in an large arc this can become a performance issue - seeing how often this function is called.

At the moment it happens sometimes, when you import units but not every cell has a value for it yet.
I try to minimize the normalizeCells calls.

Etschbeijer:
> When does this happen? What is the user story for this issue?
> 
> `let normalizeCells ` is called at many points in the code. If we ever forget calling it at a similiar position we run into the same error again. This looks like ducktaping the actual issue. Is this null issue an error in the json or the spreadsheet parsing? Depending on that we must fix this issue at the appropatie reading function.
> 
> In addition `normalizeCells` copies the table with fresh cells in an large arc this can become a performance issue - seeing how often this function is called.

I removed all normalizeCells calls that are not required. Only those that surpress this issue remain.

Freymaurer:
in: src\Server\Server.fs

```fsharp
let templateApi credentials =
    let templateUrl =
        @"https://github.com/nfdi4plants/Swate-templates/releases/download/latest/templates_v3.0.0.json"

    {
        getTemplates =
            fun () -> async {
                let! templates = ARCtrl.Template.Web.getTemplates (Some templateUrl)
                let templatesJson = ARCtrl.Json.Templates.toJsonString 0 (Array.ofSeq templates)
                return templatesJson
            }

        // ..
    }
```

But can only be fixed when release swate image too. So in the next week. @Etschbeijer i think we can close this PR? Can you test if using the new template json works as intended?

PR #1125, opened 2026-06-10
Feature/improve account manager

Freymaurer:
- Add warning icon if the currently active account has a invalid token
- regenerate link also changes ui to display input mask with the original account url, also removes account.
  - This still needs some additional improvement, but i am not exactly sure what this should do, and we are somewhat in a hurry

---

image

---


Etschbeijer (approved):
lgtm!

PR #1128, opened 2026-06-11
Start working on json import/export widget

Freymaurer:
image


src/Components/src/Composite/Widgets/JsonExport/JsonExport.fs
```diff
@@ -0,0 +1,100 @@
...
+        let exportJson =
+            defaultArg onExportJson (fun (arcFile, jsonFormat) -> defaultExportJson (arcFile, jsonFormat))
+
+        let supportedFormats =
+            React.useMemo (
+                (fun () -> Json.Generic.supportedExportFormats arcFile.RelatedArcFilesDiscriminate),
+                [| box arcFile.RelatedArcFilesDiscriminate |]
+            )
+
+        React.useEffect (
```
Freymaurer: Why not check this during set of the state variable?

useEffect should be avoided whenever possible

src/Components/src/Composite/Widgets/JsonFormatSelect.fs
```diff
@@ -0,0 +1,52 @@
+namespace Swate.Components.Composite.Widgets
+
+open Fable.Core
+open Feliz
+open Swate.Components.Shared
+
+module JsonWidgetLayout =
```
Freymaurer: Must be private module. No exports allowed except for ReactComponents

src/Components/src/Composite/Widgets/JsonImport/ImportModal.fs
```diff
@@ -0,0 +1,292 @@
...
+open System
+open Fable.Core
+open Fable.Core.JsInterop
+open Feliz
+open ARCtrl
+open Swate.Components.Composite.Template
+open Swate.Components.Composite.Template.Types
+open Swate.Components.Primitive.BaseModal
+
+module private JsonImportHelper =
```
Freymaurer: JsonImportModalHelper. This might have been a mistake i have done

Freymaurer: This can lead to build errors, as we so have the same modal name someqhere else

src/Components/src/Composite/Widgets/JsonImport/JsonImport.fs
```diff
@@ -0,0 +1,168 @@
...
+        let importJson =
+            defaultArg onImportJson (fun request -> defaultImportJson arcFile setArcFile request)
+
+        let fileType = arcFile.RelatedArcFilesDiscriminate
+
+        let supportedFormats =
+            React.useMemo ((fun () -> Json.Generic.supportedImportFormats fileType), [| box fileType |])
+
+        React.useEffect (
+            (fun () ->
```
Freymaurer: Same as above. Avoid useEffect, just habdle this during select. Or disable some of the options so this case cannot appear

src/Components/src/Page/ArcFileEditor/ArcFileEditor.fs
```diff
@@ -61,6 +62,41 @@ type private LazyComponents =
             onPickPaths = onPickPaths
         )
 
+    [<ReactComponent>]
```
Freymaurer: This is not lazy and defeats the purpose of this wrapper.

src/Components/src/Page/ArcFileEditor/ArcFileEditor.fs
```diff
@@ -61,6 +62,41 @@ type private LazyComponents =
...
+            arcFile,
+            setArcFile,
+            ?pickJsonFile = pickJsonFile,
+            ?onImportJson = onImportJson,
+            onError = onError
+        )
+
+
+    [<ReactComponent>]
+    static member LazyJsonExportWidget
```
Freymaurer: This is not lazy and cannot be lazy, please check reference on feliz docs. It must only be exactly the component which should be dynamically imported

src/Components/src/Page/ArcFileEditor/ArcFileEditor.fs
```diff
@@ -259,6 +297,9 @@ type Main =
             pickPaths: unit -> Fable.Core.JS.Promise<string[]>,
             ?trailingNavbarElements: ArcFileEditorHeaderProps -> ReactElement,
             ?startingActiveView: ActiveView,
+            ?pickJsonFile: unit -> JS.Promise<Result<JsonImportFile option, exn>>,
```
Freymaurer: Did these not work? Must we pass them from somewhere else? What is the reason behind this?

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Freymaurer, line 1: Not sure if these were required due to some funcrionality. But generic browser logic was good enough for this featureset.

Abd these are hyper specific functions, could they not have been "pickFile","saveFile" with some optional options to set dialog title and stuff? I think we already have the generic ones

src/Shared/ARCtrl.Helper.fs
```diff
@@ -292,6 +341,103 @@ module Json =
 
     module Import =
 
+        let private arcFileTypeLabel (fileType: ArcFilesDiscriminate) =
+            match fileType with
+            | ArcFilesDiscriminate.Investigation -> "Investigation"
```
Freymaurer: ArcFilesDiscriminate is a StringEnum? Therefore in js just a string? You can just call `fileType |> unbox<string>`

src/Shared/ARCtrl.Helper.fs
```diff
@@ -292,6 +341,103 @@ module Json =
...
+
+                if usedNames.Contains candidate then
+                    loop (index + 1)
+                else
+                    candidate
+
+            loop 0
+
+        let private copyTableWithName (name: string) (table: ArcTable) =
+            let copiedTable = ArcTable.init name
```
Freymaurer: Is this not: table.Copy() and then set new name?

src/Shared/ARCtrl.Helper.fs
```diff
@@ -292,6 +341,103 @@ module Json =
...
+
+                        for sourceTable in importedTables do
+                            let nextName = uniqueTableName usedNames sourceTable.Name
+                            targetTables.AddTable(copyTableWithName nextName sourceTable)
+                            usedNames <- usedNames.Add nextName
+
+                        Ok(ArcFiles.refreshRef currentArcFile)
+                | _ -> Ok importedArcFile
+
+        let tryParseToArcFile
```
Freymaurer: This is so close to the function below? Why create yet another function in a helper namespace which does nearly the same? Might be better to just use this variant whereever required

Freymaurer (commented):
Needs so.e changes. A lot of cleanup, and major issues about the missing lazy components

src/Shared/ARCtrl.Helper.fs
```diff
@@ -292,6 +334,65 @@ module Json =
...
+                let candidate = if index = 0 then baseName else $"{baseName} {index}"
+
+                if usedNames.Contains candidate then
+                    loop (index + 1)
+                else
+                    candidate
+
+            loop 0
+
+        let private copyTableWithName (name: string) (table: ArcTable) =
```
Freymaurer, line 356:
```fsharp
open ARCtrl

let table = ArcTable("My Cool Table")

let private copyTableWithName (name: string) (table: ArcTable) =
  let next = table.Copy()
  if next.Name <> name then
    ARCtrl.IdentifierSetters.setArcTableName name next
  else
    next

```

Yeah, the internal identifier setters are a bit hidden this should be improved. But here you have the easiest and most performant way to copy it.

Freymaurer (commented):
One mini change then MERGE!

src/Shared/ARCtrl.Helper.fs
Etschbeijer: fixed with commit: 445d944eedb80b766ab67d5008326b0cc3271867

PR #1135, opened 2026-06-12
Feature/fix git no remote

caroott:
Adds a safer DataHub publish flow for local ARCs that do not yet have a configured Git remote. When the default repository name already exists on the remote, the user can rename the open ARC/project and retry the push.

## Changes

- Supports pushing when `origin` is missing or has no usable URL by creating/configuring the DataHub remote before the push.
- Adds token-backed Git remote setup for publish/push operations.
- Detects remote repository name conflicts during first publish.
- Adds a publish-rename prompt in the Git sidebar.
- Renames the active ARC root on disk and updates the in-memory ARC path before retrying the push.
- Reuses the shared rename-name validation rules for ARC root rename planning.
- Clears pending file watcher state during ARC root rename so stale watcher events do not affect the renamed in-memory ARC.
- Handles renderer/main-process race conditions where `pathChange` may arrive before publish rename completion.
- Keeps post-rename side effects non-fatal after a successful disk move, avoiding split state between main process, renderer, and recent ARC tracking.

### DIFF
--- a/src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
@@ -13,6 +13,7 @@ open Main
 open Main.ARCtrlExtensions
 open Main.Bindings
 open Node.Api
+open Swate.Electron.Shared.RenamePathRules
 
 let private fsPromisesDynamic: obj = importAll "fs/promises"
 let private pathDynamic: obj = importAll "path"
@@ -169,15 +170,6 @@ type OpenArcRootRenamePlan = {
     NewName: string
 }
 
-let private normalizeArcRootFolderName (newName: string) =
-    newName
-    |> Option.ofObj
-    |> Option.map _.Trim()
-    |> Option.defaultValue String.Empty
-
-let private containsPathSeparator (value: string) =
-    value.IndexOf("/", StringComparison.Ordinal) >= 0 || value.IndexOf("\\", StringComparison.Ordinal) >= 0
-
 let private resolveAbsolutePath (pathValue: string) =
     pathDynamic?resolve (pathValue) |> unbox<string> |> PathHelpers.normalizePath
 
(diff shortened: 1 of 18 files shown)


src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
```diff
@@ -159,6 +164,138 @@ let swatelogfn id fmt =
...
+type OpenArcRootRenamePlan = {
+    SourcePath: string
+    TargetPath: string
+    NewName: string
+}
+
+let private resolveAbsolutePath (pathValue: string) =
+    pathDynamic?resolve (pathValue) |> unbox<string> |> PathHelpers.normalizePath
+
+let private normalizePathForComparison (pathValue: string) =
```
Etschbeijer: Do not pack this in an extra function please. Just call PathHelpers.normalizePathForFsComparison where it is needed.

src/Electron/src/Main/Git/GitTokenProvider.fs
```diff
@@ -23,6 +23,27 @@ let setTokenProvider (provider: GitTokenProvider) = activeTokenProvider <- provi
...
+                return Error "No usable DataHub account is signed in. Sign in before publishing this local repository."
+            }
+    }
+
+    let mutable private activeProvider = defaultProvider
+
+    let setProvider (provider: Provider) =
+        activeProvider <- provider
+
+    let createProject (projectName: string) =
```
Etschbeijer, line 44: Call activeProvider.CreateProject directly please and do not pack it into another function, please

Etschbeijer (changes requested):

src/Electron/src/Main/Git/GitTokenProvider.fs
caroott: This fits better in this module style (same as set)

src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
caroott: [7cf446b](https://github.com/nfdi4plants/Swate/pull/1135/commits/7cf446b986c8c8fa6326944ab52ec01c132d191d)

PR #1138, opened 2026-06-13
Reverted last PR and adapted GitService

Etschbeijer:


src/Components/src/Page/FileExplorer/Helper.fs
```diff
@@ -2,6 +2,14 @@ module Swate.Components.Page.FileExplorer.Helper
 
 open Swate.Components.Page.FileExplorer.Types
 
+let isLfs (item: FileItem) = item.IsLFS = Some true
+
+let needsLfsDownload (item: FileItem) =
+    isLfs item && (item.Downloaded <> Some true || item.IsLFSPointer = Some true)
+
+let hasLocalLfsCopy (item: FileItem) =
+    isLfs item && item.Downloaded = Some true && item.IsLFSPointer <> Some true
```
caroott: `item.IsLFS = Some true & item.Downloaded = Some true & item.IsLFSPointer None` would cause this to return true, even if it wasn't the file. But I'm not sure if that can happen.

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -225,34 +252,9 @@ let buildFetchRefetchArgs (relativePath: string) = [|
...
-        with
-        | None -> Error "The file is not listed by Git LFS in the current checkout."
-        | Some file ->
-            Ok {
-                file with
-                    name = PathHelpers.normalizeSeparators file.name
-            }
-    with error ->
-        Error $"Could not read Git LFS file metadata: {error.Message}"
+let buildCheckoutArgs (relativePath: string) = [| "lfs"; "checkout"; relativePath |]
```
caroott: This is missing the pathspec delimiter (see `runGitLfs`). This can cause valid filenames starting with a `-` to be interpreted as flags. Use `[| "lfs"; "checkout"; "--"; relativePath |]`

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
```diff
@@ -143,16 +156,13 @@ type FileTree =
...
 
-                    match result with
-                    | Ok pageState ->
-                        console.log ("[Renderer] Received data, processing...")
-                        pageStateCtx.setState (Some pageState)
-                    | Error errorMessage ->
-                        let fullErrorMessage = $"Could not open preview for '{item.Name}': {errorMessage}"
-                        console.log ($"[Renderer] Error: {fullErrorMessage}")
-                        pageStateCtx.setState (Some(Renderer.Types.PageState.ErrorPage fullErrorMessage))
+                    if Swate.Components.Page.FileExplorer.Helper.needsLfsDownload item then
```
caroott, line 160: downloading an LFS pointer can leave the main panel stale. The IPC action refreshes the file tree on success at IGitApi.fs line 416, but selected preview reloads exclude PageState.GitLfsPointerPage in FileExplorerDeleteHelper.fs line 34. So if the pointer page is open and the user downloads via the sidebar pill/menu, the tree updates but the main content still shows the pointer download page. Include GitLfsPointerPage _ in reloadable states or have that download action open the selected file after success.

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -1802,105 +1811,186 @@ let private discardPathspecsWithOriginals (arcPath: string) (status: StatusResul
 
     Array.append safePathSpecs originalPaths |> Array.distinct
 
-let freeLocalLfsCopy (arcPath: string) (requestedPath: string) : JS.Promise<GitResult<unit>> = promise {
-    match tryResolveArcRelativePath arcPath requestedPath with
-    | Error validationError -> return errorResult validationError
-    | Ok(safePath, absolutePath) ->
-        if not (isTrackedByAttributes arcPath safePath) then
```
caroott, line 1953: You removed the guard that checks, if something is written in the .gitattributes as far as i can see.
Now it relies on the LFS listing from GitLfsService.fs. That means a file can be “listed by Git LFS” but no longer tracked by current .gitattributes, and the free operation may still replace the local file with a pointer. That can be surprising or unsafe because later checkout/smudge behavior may not restore it the way the user expects.

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -1802,105 +1811,186 @@ let private discardPathspecsWithOriginals (arcPath: string) (status: StatusResul
...
+        (fun git -> promise {
+            let! _ = git.raw (GitLfsService.buildPullIncludeArgs safePath)
+            let! _ = git.raw (GitLfsService.buildCheckoutArgs safePath)
+            do! requireDownloadedLfsFile arcPath safePath absolutePath listing git
+            return ()
+        })
+
+let freeLocalLfsCopy (arcPath: string) (requestedPath: string) : JS.Promise<GitResult<unit>> = promise {
+    let! lfsFileResult =
+        getCleanLfsFileForAction
```
caroott, line 2084: This calls `tryGetLsFilesByRelativePath` after a chain of functions. If `git lfs ls-files -j` fails, times out, or returns unparsable JSON, free-local-copy aborts, but reports it as “not listed by Git LFS” instead of “could not read Git LFS metadata.”

caroott (changes requested):

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: Fixed in commit: 5f5a0b917dc36074537be4107d30aa99222f2ba3

src/Components/src/Page/FileExplorer/Helper.fs
Etschbeijer: Fixed in commit: 5f5a0b917dc36074537be4107d30aa99222f2ba3

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
Etschbeijer: Fixed in commit: 5f5a0b917dc36074537be4107d30aa99222f2ba3

src/Electron/src/Main/Git/GitService.fs
Etschbeijer: Fixed in commit: 5f5a0b917dc36074537be4107d30aa99222f2ba3

src/Electron/src/Main/Git/GitService.fs
Etschbeijer: Fixed in commit: 5f5a0b917dc36074537be4107d30aa99222f2ba3

src/Electron/src/Main/Git/GitLfsService.fs
```diff
@@ -225,34 +298,9 @@ let buildFetchRefetchArgs (relativePath: string) = [|
...
-    "-l"
-    "--size"
-    "--json"
-    $"--include={relativePath}"
-|]
-
-let tryFindListingForPath (relativePath: string) (listingJson: string) : Result<GitLfsLsFileInfo, string> =
-    try
-        let normalizedPath = PathHelpers.normalizeSeparators relativePath
+let buildPullIncludeArgs (relativePath: string) = [| "lfs"; "pull"; $"--include={relativePath}" |]
```
caroott: comma-path LFS downloads are still not safe when the object is actually missing locally. `downloadMissingLfsFile` fetches with this command. You correctly moved metadata lookup away from `--include`, but fetch still uses it (see GitService.fs line 2052). `git lfs pull --include=data/a,b` would leave the file as a pointer for example. You could fetch by OID from the listing, then do `git lfs checkout -- <path>` for example

src/Electron/src/Main/Git/GitService.fs
```diff
@@ -1946,105 +1957,191 @@ let private discardPathspecsWithOriginals (arcPath: string) (status: StatusResul
...
-                                                            $"Could not replace '{safePath}' with an LFS pointer without changing Git status."
-                                                    )
-                                            | Ok _ ->
-                                                if existsSync backupPath then
-                                                    unlinkSync backupPath
+                    | Ok _ ->
+                        let backupPath = createTemporaryLfsBackupPath absolutePath
+
+                        try
+                            renameSync absolutePath backupPath
```
caroott, line 2109: Only a cosmetic concern:
This moves the full local file out of the way. Then this checks out the HEAD version with smudge disabled. IPC refreshes the file tree after success, and the renderer may reload the selected path into the main panel. Since the file is now a pointer, the main panel can show raw pointer text. This bypasses the manual-open guard that clears the page for needsLfsDownload.

caroott (approved):
lgtm

src/Electron/src/Main/Git/GitLfsService.fs
Etschbeijer: fixed in comment: 173d24ac534028e8735c6baf6dd3350fdc16db58

src/Electron/src/Main/Git/GitService.fs
Etschbeijer: fixed in comment: 173d24ac534028e8735c6baf6dd3350fdc16db58

PR #1159, opened 2026-06-15
Fix git environment for mac/linux and set lfs download to false by default

caroott:
This PR adds path context to the simple git save environment for mac/linux compatibility and sets default lfs download to false again.


Etschbeijer (approved):
Lgtm

PR #1168, opened 2026-06-15
preview window full height

Rookabu:
Preview window is now full height and should grow/shrink if the window size changes.

Edit:
grafik
Preview:
grafik

### DIFF
--- a/src/Components/src/Composite/MarkdownText/TextInputWithMarkdown.fs
+++ b/src/Components/src/Composite/MarkdownText/TextInputWithMarkdown.fs
@@ -82,7 +82,8 @@ type TextInputWithMarkdown =
 
         let options = {
             MarkdownOptions.defaults with
-                Height = defaultArg height MarkdownOptions.defaults.Height
+                MinHeight = defaultArg height MarkdownOptions.defaults.MinHeight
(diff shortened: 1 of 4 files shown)


src/Components/src/Composite/MarkdownText/TextInputWithMarkdown.fs
```diff
@@ -423,8 +424,8 @@ type TextInputWithMarkdown =
 
         let previewClassName =
             match options.PreviewClassName with
-            | Some className -> $"swt:p-4 {className}"
-            | None -> "swt:p-4"
+            | Some className -> $"swt:p-4 swt:h-[360px]{className}"
```
caroott:
```
prop.style [
    style.height options.MaxHeight
    style.minHeight options.MinHeight
]
```
 already sets style.maxHeight options.MaxHeight (360 default) and style.minHeight options.MinHeight (240 default) on the wrapper `<div>` that contains MarkdownPreview. Adding `swt:h-[360px]` should be unnecessary.

caroott (changes requested):

Etschbeijer (approved):
Lgtm

PR #1175, opened 2026-06-15
Git initialization flow and ARC creation

caroott:
Three interconnected changes to the Git sidebar initialization flow and ARC creation:

1. **Remove the direct-publish-through-namefield option from the Git sidebar init screen**
2. **Add a default-checked "Initialize Git Repository" checkbox to ARC creation**
3. **Set the default branch name to `main` instead of `master` on `git init`**

---

### 1. Remove old repo initialization path (`bb818be6`)

**Files:** `GitSidebarPanel.fs`, `GitWorkflow.fs`, `GitStateContext.fs`, `GitWorkflow.test.fs`

Removed the "DataHub repository name" text field from the missing-repository state in the Git sidebar. Previously, entering a name would create a DataHub project and set it as `origin` remote during init. This was a separate feature from the publish-on-push flow (which remains untouched).

- Removed `remoteProjectName` state hook, `normalizedRemoteProjectName` computation, and the `extraContent` UI block (text input + help text + `?infoText`)
- Changed `InitRepositoryRequested of remoteProjectName: string option` to parameterless `InitRepositoryRequested`
- Removed `createDataHubProject` and `gitAddRemote` from `GitDependencies`
- Simplified `runInitRepositoryAsync` to only do local `git init` — no DataHub project creation or remote setup
- Removed unused `open Swate.Components.Api.GitLabApi` (the publish-on-push flow calls `GitLabApi.CreateProject` directly in the main process, not through this dependency chain)
- Updated `GitStateController.initRepository` from `string option -> unit` to `unit -> unit`
- Removed two obsolete tests and updated the remaining init test

### 2. Add init git by default to ARC creation (`83162f41`)

**Files:** `InitState.fs`, `ArcVaultHelper.fs`, `IPCTypes.fs`, `IArcVaultsApi.fs`, `ArcVault.fs`

Added a checkbox "Initialize Git Repository" (checked by default) to the ARC creation modal. When checked, the new ARC gets a git repository initialized automatically — using the same `git init` flow as the sidebar. The modal is shared between the init-state page and the navbar selector, so both entry points benefit.

**IPC contract change:** `createARC` now accepts a `CreateArcRequest` record with `identifier: string` and `initGit: bool` instead of a plain string. This threads the flag through the main process call chain (`CreateOrFocusArc` → `RegisterVaultWithNewArc` / `CreateARC`).

### 3. Rename master to main on git init (`023503ba`)

**File:** `GitProvisioningService.fs`

Changed `git.init()` to `git.init (U2.Case1 [| "--initial-branch=main" |])`. Compatible with git >= 2.28.

### DIFF
--- a/src/Electron/src/Main/ArcVault/ArcVault.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVault.fs
@@ -387,7 +387,7 @@ module ArcVaultExtensions =
                 sendMsg.pathChange (Some normalizedPath)
         }
 
-        member this.CreateARC(path: string, identifier: string, initGit: bool) = promise {
+        member this.CreateARC(path: string, identifier: string) = promise {
             match this.path, this.arc with
             | Some _, _ -> swatefailfn this.window.id "Unable to create ARC in vault bound to path."
             | _, Some _ -> swatefailfn this.window.id "Unable to create ARC in vault bound to ARC."
@@ -414,11 +414,6 @@ module ArcVaultExtensions =
                             (PathHelpers.formatContractErrors errors)
                 finally
                     this.isBusyWriting <- false
-
-                if initGit then
-                    match! Git.GitProvisioningService.initRepository normalizedPath with
-                    | Error failure -> swatelogfn this.window.id $"Git init failed for '{normalizedPath}': {failure}"
-                    | Ok _ -> ()
 
                 do! this.Startup()
                 sendMsg.pathChange (Some normalizedPath)
@@ -511,6 +506,14 @@ type ArcOpenDisposition =
     | FocusedExisting of path: string
     | CreatedInCurrent of path: string
     | CreatedInNewWindow of path: string
+
+    member this.CreatedArcPath =
+        match this with
+        | CreatedInCurrent path
+        | CreatedInNewWindow path -> Some path
+        | OpenedInCurrent _
+        | OpenedInNewWindow _
+        | FocusedExisting _ -> None
 
 module ArcOpenDisposition =
     let path =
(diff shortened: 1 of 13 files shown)


src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -415,6 +415,11 @@ module ArcVaultExtensions =
                 finally
                     this.isBusyWriting <- false
 
+                if initGit then
```
Freymaurer: Please do not call any git functionalty in this file. Keep domain specific. Call git init in `ipc/IArcVaultsAPI.fs`. 

Remember to not use window path, but path of returned arc.create function or result from path picker.

src/Electron/src/Renderer/Components/InitState.fs
```diff
@@ -60,11 +71,32 @@ let CreateNewArcModalContent (close: unit -> unit) =
                 ]
             ]
         ]
+        Html.div [
+            prop.className "swt:form-control swt:mt-4"
```
Freymaurer: form-control is deprecated in daisyui 4

Freymaurer: https://daisyui.com/components/checkbox/#with-fieldset-and-label

Freymaurer (changes requested):

src/Electron/src/Renderer/Components/InitState.fs
caroott: [7e2ba1d](https://github.com/nfdi4plants/Swate/pull/1175/commits/7e2ba1d03fa3843e8a0cdf99189a0fa32a899cb0)

src/Electron/src/Main/ArcVault/ArcVault.fs
caroott: [f7d362b](https://github.com/nfdi4plants/Swate/pull/1175/commits/f7d362b6440c89681f50937569ae02aab26ab107)

Etschbeijer (approved):
Lgtm

PR #1177, opened 2026-06-16
Use ErrorModal for Git Services

caroott:
- Route Git workflow errors through the shared `ErrorModal`
- Replace Git version-check alerts with modal errors
- Add renderer tests for Git error-modal reporting


Etschbeijer (approved):
Lgtm

Issue #1145, opened 2026-06-15
[Bug]: Swate App: Renaming/deleting of dataset/protocol folder


SabrinaZander:
### What happened?

Would not allow renaming  or deleting of dataset and protocol folders as it could probably mess up with the ARC structure. 

<img width="259" height="128" alt="Image" src="https://github.com/user-attachments/assets/cf487d58-0c49-4137-a299-c9a39227325d" />

### Host

Applies to All

SabrinaZander:
After commiting and uploading, the default folders are added again + renamed one.

<img width="261" height="289" alt="Image" src="https://github.com/user-attachments/assets/b3b6020c-39de-4e4d-a9e9-a78f166357c8" />

PR #1185, opened 2026-06-17
Remove delete and rename button from datasets and standard children

Etschbeijer:
Closes issue #1145


src/Shared/PathHelpers.fs
```diff
@@ -148,8 +148,15 @@ module ArcEntityPathRules =
...
 
+    type private StructuralArcPath =
+        | AddZoneRoot
+        | EntityFolder
+        | ProtectedEntityChildFolder
+        | OtherPath
+
     let private protectedDeleteTargetNames = [ ".gitattributes"; ".gitkeep"; "readme.md" ]
     let private protectedRenameRootFolderNames = [ "notes" ]
+    let private protectedEntityChildFolderNames = [ "dataset"; "protocol"; "protocols" ]
```
caroott: i think it's only protocols, not both. Also, resources should be protected as well then.

tests/Electron.Renderer/FileTreeRenameWorkflow.test.fs
```diff
@@ -194,6 +204,23 @@ Vitest.describe (
                 )
         )
 
+        Vitest.test (
+            "rename context menu item is hidden for protected structural entity child folders",
+            fun () ->
+                let protectedChildFolderPaths = [
+                    "assays/AssayA/dataset"
+                    "studies/StudyA/protocol"
+                    "workflows/WorkflowA/protocols"
```
caroott: workflows don't have a protocols folder. if yes, it should be renamable/removable. Same for dataset in runs.

tests/Shared/PathHelpers.Tests.fs
```diff
@@ -155,6 +155,18 @@ let tests =
                 (ArcEntityPathRules.isDeletePathAllowed "assays/MyAssay/.gitattributes")
                 ".gitattributes should remain non-deletable."
 
+            Expect.isFalse
+                (ArcEntityPathRules.isDeletePathAllowed "assays/MyAssay/dataset")
+                "Structural dataset folders should remain non-deletable."
+
+            Expect.isFalse
+                (ArcEntityPathRules.isDeletePathAllowed "studies/MyStudy/Protocols")
```
caroott: do you check case sensitivity? If yes, they should be lowercase. if no, then you could run into trouble with linux os.

caroott (changes requested):

src/Shared/PathHelpers.fs
Etschbeijer: Fixed in commit: 3f9e3ef94250ad443b05e05c04a0ff30f412eac8

tests/Electron.Renderer/FileTreeRenameWorkflow.test.fs
Etschbeijer: Fixed in commit: 3f9e3ef94250ad443b05e05c04a0ff30f412eac8

tests/Shared/PathHelpers.Tests.fs
Etschbeijer: Fixed in commit: 3f9e3ef94250ad443b05e05c04a0ff30f412eac8

caroott (approved):

Issue #1183, opened 2026-06-17
[Bug]: Initiating a note creates always an empty asset folder


Etschbeijer:
### What happened?

I created a note and an empty asset folder was initialized.

### Host

Electron Swate App

### Host Details

Windows

Thyra:
I think this is not a bug but desired behavior, otherwise the risk that users put additional files (PDFs, images, whatever they can think of) just into the note root folder or some other random subfolder is too high. I would rather codify the asset subfolder as part of the specification. We could add a .gitkeep file to include them in git tracking?

Issue #1182, opened 2026-06-17
[Feature Request]: Note asset handling


Thyra:
### Describe the solution you would like!

Now with the new structure for notes, we should
- have the asset files marked as LFS files
- make sure when we attach/move a note to a study or assay, the asset files (and I guess any other files in the notes folder) are moved to resources/dataset respectively.

### Host

Electron Swate App

Thyra:
Even better: when moving a note to a study or assay, allow the user to select which assets should be moved into dataset/resources and which should move into the protocols folder (or additional payload) with the MD file

PR #1187, opened 2026-06-18
Add further asset handling for notes

Etschbeijer:
* Initialize asset folder only when an image is existing
* Give assay or study the ability to assign a note after its creation
* Move asset folder when a note is assigned to a study or assay afterward
* Move note related logic out of filetree into FileTreeAssignNoteHelper.fs
* When a target already exists stop the assignment
* Fix error so that multiple images can be added at once


Etschbeijer:
Closes issue #1182
Closes issue #1183

Etschbeijer:
Select study and assay modal.
grafik

Etschbeijer:
Overwrite modal.
grafik

Etschbeijer:
New button in context menu.
grafik

Etschbeijer:
Assign note to assay / study modal

grafik

Etschbeijer:
Target with the same name already exists.
grafik

caroott:
> * Move asset folder when a note is assigned to a study or assay afterward

Do you move the folder or copy the content to the protocols folder?
And what happens if an asset folder already exists in the study/assay? With the current design, we have one asset folder there for several protocols.

Etschbeijer:
> > * Move asset folder when a note is assigned to a study or assay afterward
> 
> Do you move the folder or copy the content to the protocols folder? And what happens if an asset folder already exists in the study/assay? With the current design, we have one asset folder there for several protocols.

We move it there, so it is no longer available in the original notes folder.
When no asset folder is available, it is created.
Otherwise, when an asset folder is already there, it will be used.

caroott:
> We move it there, so it is no longer available in the original notes folder. When no asset folder is available, it is created. Otherwise, when an asset folder is already there, it will be used.

We shouldn't move assets and protocols, that would make the notes folder as a lab notebook less useful and prevent reuse of existing protocols. It should always be a copy.
When we copy the protocols and asset items, we should also prefix the new name with its origin to prevent overwrites (e.g. YYYY_MM_DD_protocolName.md and YYYY_MM_DD_protocolName_imageName.png)

Etschbeijer:
> > We move it there, so it is no longer available in the original notes folder. When no asset folder is available, it is created. Otherwise, when an asset folder is already there, it will be used.
> 
> We shouldn't move assets and protocols, that would make the notes folder as a lab notebook less useful and prevent reuse of existing protocols. It should always be a copy. When we copy the protocols and asset items, we should also prefix the new name with its origin to prevent overwrites (e.g. YYYY_MM_DD_protocolName.md and YYYY_MM_DD_protocolName_imageName.png)

Fixed in commit: dd63e4ce4e87cce312cb3e2a6bb45a6c426f91a3

Etschbeijer:
Note with multiple assets
grafik

Etschbeijer:
Add header selector for assign asset target
grafik

Freymaurer:
This also scrolls the search field? And it should just scroll the search results

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -257,12 +263,40 @@ module ArcFileSystemHelper =
         | FileSystemItemKind.File -> ARCtrl.FileSystemHelper.writeFileTextAsync targetAbsolutePath ""
         | FileSystemItemKind.Folder -> mkdirAsync targetAbsolutePath
 
+    let private copyFileAsync sourceAbsolutePath targetAbsolutePath = promise {
+        let! _ =
+            fsPromisesDynamic?copyFile (sourceAbsolutePath, targetAbsolutePath)
```
Freymaurer: reuse existing js bindings. avoid dynamic operators

Freymaurer: this applies to a LOT of code below

Freymaurer: I also updated the AGENTS.md file with these requirements

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Freymaurer, line 1: Why do we want a specialized imagePaths?

Why not pickAbsolutePaths with an optional argument `string []` for filter file types?

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -612,6 +656,18 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
             with e ->
                 return Error e
         }
+    copyPath =
```
Freymaurer: What does this do? Because it uses `copyGenericFileSystemItemOnDisk`, which to me does not sound like "copyPath", but rather something akin to "copy item at path"

src/Electron/src/Main/IPC/IPCHelper.fs
```diff
@@ -35,10 +35,11 @@ module IPCHelper =
...
         : JS.Promise<Result<'T, exn>> =
         promise {
+            let wasBusyWriting = vault.isBusyWriting
             vault.isBusyWriting <- true
 
             try
                 return! operation ()
             finally
-                vault.isBusyWriting <- false
+                vault.isBusyWriting <- wasBusyWriting
```
Freymaurer: good catch!

src/Electron/src/Preload/preload.fs
```diff
@@ -1,9 +1,19 @@
 ﻿module Preload
 
+open Browser.Types
+open Fable.Core.JsInterop
+open Fable.Electron.Renderer
 open Fable.Electron.Remoting.Preload
 open Swate.Electron.Shared.IPCTypes
 open Swate.Electron.Shared.IPCTypes.MainToRendererIpc
 
+contextBridge.exposeInMainWorld (
```
Freymaurer: This breaks existing conventions! We used Fable.Electron.Remoting for ALL other IPC calls. Breaking with this convention requires a very good reason!

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/AssignNoteHelper.fs
```diff
@@ -0,0 +1,303 @@
...
+open Fable.Core
+open Swate.Components.Shared
+open Swate.Components.Composite.Notes.Editor
+open Swate.Components.Page.FileExplorer.Types
+open Swate.Components.Primitive.ErrorModal.Types
+open Swate.Electron.Shared.FileIOTypes
+open Swate.Electron.Shared.FileIOHelper
+open Renderer.Components.LeftSidebar.FileExplorer.Types
+
+module FileTreeAssignNoteHelper =
```
Freymaurer: again: if you have a single module in a file. make it a root level module.

Freymaurer (changes requested):
I really do not like that we have FileSystemHelper on really multiple project locations spread around. These should be collected in either Swate.Components or any shared project.

The only sensible exception is Electron.Main, as it also has the direct fs access functions, which only make sense for a framework with this access.

Etschbeijer:
> This also scrolls the search field? And it should just scroll the search results

Fixed in commit: de79d99f1218d826871f6e334c18bea3ac2851e2

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: Fixed with commit: 4815dea0de4a73e13a5f0895c9be28a22aa3d23e

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Yes, you are right.
It is more efficient to call the absolute paths and filter them afterward. That way the responsibility for the target interaction is put into the callers responsibility.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: It copies the file or folder located at sourceRelativePath inside the active ARC to targetRelativePath inside the same ARC.
So, it is basically copying the fileSystemItem with a relative path...

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: c6f88d5e4b4d3c3e78e18cd565a04555a9ba364d

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: c6f88d5e4b4d3c3e78e18cd565a04555a9ba364d

src/Electron/src/Preload/preload.fs
Etschbeijer: Fixed in commit: d48d0aa9aeb59732c21b4889ecbbbb38f6a65714

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/AssignNoteHelper.fs
Etschbeijer: Fixed in commit: 6490adcf57ce108fa4eeced22f7cdd89b12cb9d5

Etschbeijer:
> I really do not like that we have FileSystemHelper on really multiple project locations spread around. These should be collected in either Swate.Components or any shared project.
> 
> The only sensible exception is Electron.Main, as it also has the direct fs access functions, which only make sense for a framework with this access.

I think we should create a seperate pr and issue for this.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Freymaurer: this is not what i meant. it is more efficient and better ux to let the picker handle the filtering. BUT we can pass the filters as argument INTO the ipc call

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -346,7 +387,77 @@ module ArcFileSystemHelper =
...
+                return Ok targetPath
+            with copyError ->
+                return Error copyError
+        }
+
+    let copyExternalFilesToArcOnDisk
+        (arcPath: string)
+        (requests: CopyExternalFileRequest[])
+        : JS.Promise<Result<string[], exn>> =
+        let rec copyNext copiedPaths remainingRequests = promise {
```
Freymaurer: doing this rec can run into performance issues. I checked and you could run some of them in parallel, ai gave me back the code below. It uses the count of cpu cores to determine the number of parallel processes (worker) started. Each worker runs while index tracked in mutable is below file count and copies files. due to the `while`, it just restarts as long as there are more files.

```js
import { copyFile } from "node:fs/promises";
import os from "node:os";

const concurrency = os.cpus().length * 2;

async function parallelCopy(files: { src: string; dst: string }[]) {
  let index = 0;

  async function worker() {
    while (index < files.length) {
      const current = files[index++];
      await copyFile(current.src, current.dst);
    }
  }

  await Promise.all(
    Array.from({ length: concurrency }, worker)
  );
}
```

I think this should be better optimizied, as a instable feature can create user frustration, following the github zen:

> - It’s not fully shipped until it’s fast
> - Half measures are as bad as nothing at all

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -612,6 +676,18 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
             with e ->
                 return Error e
         }
+    copyFileSystemItem =
```
Freymaurer, line 585: what is the difference between: `copyFileSystemItem` and `copyExternalFilesToArc`? Can they be combined?

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -338,6 +372,14 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
             with e ->
                 return Error(exn $"Could not pick files: {e.Message}")
         }
+    getPathForFile =
```
Freymaurer: Not sure what this function can be used for, willmaybe come back after reading more code

src/Electron/src/Main/IPC/IPCHelper.fs
```diff
@@ -1,44 +1,40 @@
...
+    let windowId = windowIdFromIpcEvent event
+
+    match ARC_VAULTS.TryGetVault(windowId) with
+    | None -> Error(exn $"The ARC for window id {windowId} should exist")
+    | Some vault ->
+        match vault.path with
+        | Some arcPath -> Ok(vault, arcPath)
+        | None -> Error(exn "ARC is not loaded.")
+
+let withBusyWriting (vault: ArcVault) (operation: unit -> JS.Promise<Result<'T, exn>>) : JS.Promise<Result<'T, exn>> = promise {
```
Freymaurer, line 32: cool!

src/Electron/src/Renderer/Components/Helper/FileSystemHelper.fs
```diff
@@ -38,27 +63,156 @@ let ensureChildFolder
...
+        PickFiles =
+            (fun () -> promise {
+                match! pickAbsolutePaths (Some imageFileExtensions) with
+                | Ok paths -> return paths |> Array.map toPromptFile |> Array.toList
+                | Error error when error.Message = "Cancelled" -> return []
+                | Error error -> return raise error
+            })
+        ResolveMarkdownPath =
+            (fun file -> promise {
+                match! tryResolvePromptFileHostPath tryResolveBrowserFilePath file with
```
Freymaurer: Can you explain this part to me? Not sure what this is about

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/Modals/AssignNoteAssetSelector.fs
```diff
@@ -0,0 +1,202 @@
...
+            assets
+            |> Seq.map (selectedValueForAsset assetDestinations)
+            |> Seq.distinct
+            |> Seq.toList
+
+        match distinctSelectedValues with
+        | [ selectedValue ] -> selectedValue
+        | _ -> MixedHeaderValue
+
+    let destinationOptions availableDestinations includeMixedOption = [
```
Freymaurer: This return ReactElement [], instead use React.Fragment and return ReactElement, make it a static member "[<ReactComponent>]"

Freymaurer (changes requested):
FileExplorer is a "Page" component, you can add all those nice modals into the component, so that you only need to call it in electron and wire up the underlying functionality. For example:

"src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/Modals/AssignNoteModal.fs"

Stopped reading pr at that point

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: 2667aac5c3b364046454f95631510336cc9141e8

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: I combined the workflow below that can be fused but they expect different kind of input data and have different abilities to handle that.
In order to unify them, we would have a clearer signaling of the source of the data.
E.g. copyFileSystemItem can copy folders recursively and copyExternalFilesToArc can only copy external files without a recursive loop.
I can combine them but that would require a small rework of the input logic, too.

Etschbeijer: Fixed in commit: f1b05f379f70685dda510592b4ca826f9dcad20c

src/Electron/src/Main/IPC/IPCHelper.fs
Etschbeijer: Really cool or what did you do here?

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: It is used in FileSystemHelper.fs for asset handling.
It receives a Browser.Types.File
-> Then it asks Electron for the host path behind that File
-> Then it normalizes the path separators
-> Then it returns Result<string, exn>

It resolves the local absolute path for copyExternalFilesToArc.
It makes the use for drop and down and so on in the future easier because it resolves the paths for them.

src/Electron/src/Renderer/Components/Helper/FileSystemHelper.fs
Etschbeijer: It resolves the selected file to the real host path and builds the markdown-relative asset path. Then it registers a pending copy by calling addAsset.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/Modals/AssignNoteAssetSelector.fs
Etschbeijer: Fixed in commit: 16d631a49c499a1134306a9b2645f2f84d38c193

Etschbeijer:
> src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/Modals/AssignNoteModal.fs

Fixed in commit: 5fa28eb35bd3a399e6de65a7f607d91186345ae7
Moved the modals to swate component

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: Tried to fix it in commit: fd50fcae53a97a7c00fc4f9f4662ddd838e4ac85
Replaced the recursive external file copy loop with bounded parallel async workers, using CPU count capped at 16.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Freymaurer: So it does:

1. User clicks and wants to run `operationXX` via FileTree interaction
2. FileTree asks `getPathForFile` and awaits
3. Main resolves absolute path
4. FileTree receives path and triggers actual operation with absolute path `operationXX`
5. Main runs `operationXX`.

Could it be the following instead:

1. User clicks and wants to run `operationXX` via FileTree interaction
2. FileTree calls `operationXX` with relative path (with relative path bool flag?)
3. Main resolves relative path and runs `operationXX`

src/Electron/src/Main/IPC/IPCHelper.fs
Freymaurer: yes, no sarcasm, i like this!

src/Electron/src/Renderer/Components/Helper/FileSystemHelper.fs
Freymaurer: Is HostPath an absolute Path? Can this not simply be handled by Main process? What is a BrowserFilePath?

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: b2edce9aeac007c1196e6c59e2e6c351ee0a0c95

src/Electron/src/Renderer/Components/Helper/FileSystemHelper.fs
Etschbeijer: Fixed in commit: d173c010995c805e8ddbae3c02532b188fd88925

src/Components/src/Composite/Notes/Editor/NoteFormFields.fs
Freymaurer, line 1: if you adjust an existing file, please follow reactcomponent design docs. namespace, static member.

src/Components/src/Page/FileExplorer/Modals/AssignNoteModal.fs
```diff
@@ -0,0 +1,119 @@
...
+            setSelectedNote: AssignableNoteRef option -> unit,
+            availableNotes: ResizeArray<AssignableNoteRef>,
+            availableAssets: ResizeArray<AssignableNoteAssetRef>,
+            availableAssetDestinations: AssignNoteAssetDestination list,
+            assetDestinations: Map<string, AssignNoteAssetDestination>,
+            setAssetDestination: string -> AssignNoteAssetDestination option -> unit,
+            close: unit -> unit,
+            submit: unit -> unit,
+            ?isAssigning: bool
+        ) =
```
Freymaurer: Replace `isAssigning` with internal state. This can be done by making submit a promise.

```suggestion
            submit: unit -> Promise<unit>,
        ) =
```

```fsharp

// .. inside component

let isRunning, setIsRunning = React.useState(false)
// clarify that we want to disable interactions while running commands
let isDisabled = isRunning

let onSubmit = fun () ->
  promise {
    setIsRunning true
    try
      try
        do! submit()
      with
        | exn -> // error handling
    finally
       setIsRunning false
  }

```

src/Components/src/Page/FileExplorer/Types.fs
```diff
@@ -105,6 +106,22 @@ type DirectoryInteractionMode =
     | OpenOnDoubleClickCloseOnSingleClick
     | ToggleOnSingleClickSelectOnDoubleClick
 
+type AssignableNoteRef = {
```
Freymaurer, line 110: just a comment, no need to fix anything for now. But with the future plan to move outside of ISA we propably have to make this more flexible

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -307,71 +332,41 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
...
                     return Error(exn "Not exactly one path")
                 else
-                    return Ok(result.filePaths |> Array.exactlyOne)
+                    return Ok(result.filePaths |> Array.exactlyOne |> PathHelpers.normalizePath)
             with e ->
                 return Error(exn $"Could not pick directory: {e.Message}")
         }
-    pickAbsolutePaths =
-        fun () -> promise {
+    pickExternalFilePaths =
```
Freymaurer:
```suggestion
    pickFilePaths =
```

src/Electron/src/Preload/preload.fs
```diff
@@ -1,9 +1,16 @@
 ﻿module Preload
 
+open Browser.Types
 open Fable.Electron.Remoting.Preload
 open Swate.Electron.Shared.IPCTypes
 open Swate.Electron.Shared.IPCTypes.MainToRendererIpc
 
+let private fileSystemApi = {|
```
Freymaurer: Did we not have this already? Please use Fable.Electron.Remoting. Do not use another pattern, except you have a very good reason to do so. In that case there MUST be a comment with a explanation

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
```diff
@@ -67,6 +93,13 @@ type FileTree =
 
         let activeDialog, setActiveDialog = React.useState<FileTreeDialog option> None
         let isDialogBusy, setIsDialogBusy = React.useState false
+
+        let selectedAssignableNote, setSelectedAssignableNote =
+            React.useState<AssignableNoteRef option> None
+
+        let selectedAssetDestinations, setSelectedAssetDestinations =
+            React.useStateWithUpdater<Map<string, AssignNoteAssetDestination>> Map.empty
```
Freymaurer: One of these can be combined with FileTreeDialog of ... . The other can be moved inside the opened modal.

Always try to keep states as low level as possible.

src/Shared/Swate.Components.Core.fsproj
```diff
@@ -49,4 +49,8 @@
     <!-- Include the project file itself as well -->
     <Content Include="$(MSBuildThisFileFullPath)" Pack="true" PackagePath="fable/" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="JsBindings\" />
+  </ItemGroup>
```
Freymaurer, line 55: ?

Freymaurer (changes requested):

src/Components/src/Composite/Notes/Editor/NoteFormFields.fs
Etschbeijer: resolved in commit: 3bdf71a2a357ac969da8d8a28849dd9a6fd3585a

src/Components/src/Page/FileExplorer/Types.fs
Etschbeijer: Fixed in comment: a0e663eb1bacd12e5de5c105872b33b509fb92b5

src/Shared/Swate.Components.Core.fsproj
Etschbeijer: I removed them already but they were still linked here.
Deleted the link here.

src/Electron/src/Preload/preload.fs
Etschbeijer: Fixed in commit: 5bb026f2bc282d9ec69f307477c33abd63256e85

src/Shared/Swate.Components.Core.fsproj
Etschbeijer: Fixed in commit: 5bb026f2bc282d9ec69f307477c33abd63256e85

src/Components/src/Page/FileExplorer/Modals/AssignNoteModal.fs
Etschbeijer: Fixed in commit: d9711f4be46f16b51676055a735ed4a18920eb6b

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: 7edac62337cafb7d4509e0b51fe7ddc928874c34

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
Etschbeijer: Fixed in commit: 2dae44966b8b7669c12354b0b2a85701d3053da0

Also updated the states of other code.

src/Components/src/Composite/Notes/Editor/NoteFormFields.fs
Etschbeijer: Updated in commit: 2dae44966b8b7669c12354b0b2a85701d3053da0

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/AssignNoteHelper.fs
```diff
@@ -0,0 +1,284 @@
...
+    (assets: AssignableNoteAssetRef list)
+    (assetDestinations: Map<string, AssignNoteAssetDestination>)
+    : JS.Promise<unit> =
+    let targetFolderPath = buildAssignedNoteFolderPath target note.NoteFolderName
+
+    promise {
+        if PathHelpers.pathsEqual note.SourceFolderPath targetFolderPath then
+            config.closeDialog ()
+        else
+            let! copyResult =
```
caroott, line 317: You now copy it and don't move, but you still do not prefix it with the date, right? This way you still overwrite notes with the same name.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/AssignNoteHelper.fs
```diff
@@ -0,0 +1,284 @@
...
+                config.copyFileSystemItem {
+                    sourceRelativePath = note.SourceFolderPath
+                    targetRelativePath = targetFolderPath
+                    overwrite = true
+                }
+
+            match copyResult with
+            | Error copyError -> enqueueAssignNoteError config.enqueueError copyError.Message
+            | Ok() ->
+                match! moveAssignedAssets config target note targetFolderPath assets assetDestinations with
```
caroott: if the note copy succeeds but moveAssignedAssets fails midway, you're left with a half-copied note (folder present, some assets moved, some not) and an error

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -263,35 +255,59 @@ module ArcFileSystemHelper =
...
+                    sourceAbsolutePath
+                    targetAbsolutePath
+                    (Main.Bindings.Filesystem.CpOptions(recursive = true, force = false))
+
+            return Ok()
+        with copyError ->
+            return Error(exn $"Cannot copy '{sourcePath}' to '{targetPath}': {copyError.Message}")
+    }
+
     ///WIP must be simplified in future pr
```
caroott, line 289: Do we have an issue for this?

caroott (changes requested):

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: No, a comment for myself.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/AssignNoteHelper.fs
Etschbeijer: You are right! I forgot to prefix it with the date, thank you!

Etschbeijer: Fixed in commit: 6067d0ec44b7ebff836e2235725fa7ec75292bc7

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/AssignNoteHelper.fs
Etschbeijer: Fixed in commit: c2fb461fd73a1aa578f0da7302dfed4dea976dfd

src/Electron/src/Main/IPC/FileSystemIO.fs
caroott: Then i would do it in this PR or write the issue

Etschbeijer: Created the issue: https://github.com/nfdi4plants/Swate/issues/1241

Issue #1163, opened 2026-06-15
[Feature Request]: Add default .gitignore


Brilator:
### Describe the solution you would like!

e.g. https://github.com/nfdi4plants/ARCCommander/blob/b5a330dbf642f0a55c7aa15a9a60a057d058a30f/src/ArcCommander/defaultGitignore

### Host

Electron Swate App

PR #1189, opened 2026-06-19
Add basic gitignore

Etschbeijer:
Add basic gitignore, based on gitignore of ARCtrl


Etschbeijer:
Closes issue #1163

src/Electron/src/Main/ArcScaffold.fs
```diff
@@ -0,0 +1,43 @@
+namespace Main
+
+open ARCtrl.Contract
+
+module ArcScaffold =
+
+    let defaultGitignoreContent =
+        ("""# ARC local/generated artifacts
+/arc-validate-results.xml
+/.cwl/
```
caroott: why do we want to hide workflow files?

Etschbeijer: Fixed in commit: d7261ef7ea75b261a3b8c2b23ab45c83978532d5

caroott:
Do you have a source for the choices in this .gitignore?
I know that github offers a selection of templates for different use cases:
image

Is it one of those or a custom one?
Some of those files i've never seen in an ARC, even locally.

src/Electron/src/Main/ArcScaffold.fs
```diff
@@ -0,0 +1,9 @@
+namespace Main
+
+open ARCtrl.Contract
+
+module ArcScaffold =
```
Freymaurer: if you plan to create a exactly one module with the name of the file, alway do a root level module. 

`module Main.ArcScaffold` at the top of the file

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -409,7 +409,13 @@ module ArcVaultExtensions =
 
                 try
                     match! arc.TryWriteAsyncSwate(normalizedPath) with
-                    | Ok _ -> ()
```
Freymaurer, line 412: to my knowledge, `arc.TryWriteAsyncSwate` is a custom function we use for WriteAsync. Can we not inject the gitignore contract in there? And open an issue in ARCtrl to ask for support of "Add custom contracts to TryWriteAsync?"

Then we can actually inject the notes folder creation contract, as well as the gitignore contract into the same write command. This has the advantage that it looks more fluid. Currently in a new ARC the notes folder appears like 2 seconds after the rest, which is not optimal

Freymaurer (changes requested):

src/Electron/src/Main/ArcVault/ArcVault.fs
Freymaurer: image

Freymaurer (approved):
lgtm!

Issue #1169, opened 2026-06-15
[Feature Request]: Improve visibility of annotation table


SabrinaZander:
### Is your feature request related to a problem?

For me still relevant in new Swate App

https://github.com/nfdi4plants/ARCitect/issues/391

### Host

Applies to All

PR #1190, opened 2026-06-19
Improve annotationtable visibility in swate

Etschbeijer:
* Added table creation after initialization of study, assay and run
* Applied Fantomas formatting

### DIFF
--- a/src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
+++ b/src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
@@ -135,6 +135,16 @@ module ArcLoadExtensions =
         return repairedAny
     }
 
+    let private ensureDefaultAnnotationTables (arc: ARC) =
+        for study in arc.Studies do
+            (ArcFiles.Study(study, [])).EnsureDefaultAnnotationTable() |> ignore
+
+        for assay in arc.Assays do
+            (ArcFiles.Assay assay).EnsureDefaultAnnotationTable() |> ignore
+
+        for run in arc.Runs do
+            (ArcFiles.Run run).EnsureDefaultAnnotationTable() |> ignore
+
     type ARC with
 
         /// Hotfix for #619, not fixed in the consumed ARCtrl 3.0.0-beta.12.
@@ -147,6 +157,7 @@ module ArcLoadExtensions =
             match! fullFillContractBatchAsync arcPath contracts with
             | Ok fulfilledContracts ->
                 arc.SetISAFromContracts fulfilledContracts
+                ensureDefaultAnnotationTables arc
                 return Ok arc
             | Error errors -> return Error errors
         }
--- a/src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
+++ b/src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
@@ -3,8 +3,10 @@ namespace Main.ARCtrlExtensions
 open System
 open ARCtrl
 open ARCtrl.Contract
-open Main.Bindings.Filesystem
 open Main.Bindings.Path
+open Main.Bindings.Filesystem
+open Main.ArcMerge.ArcFileDefaults
+open Swate.Components.Shared
 open Swate.Electron.Shared.FileIOHelper
 
 [<AutoOpen>]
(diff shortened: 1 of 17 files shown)


Etschbeijer:
Closes issue #1169

Freymaurer (changes requested):
Why do we do this? What is the advantage? From the code i saw the new logic is:

Create new ArcFile -> Start with table, which name is the identifier of the ArcFile and has a single input column.

If this is true, why is this better than no table or an empty table, which would trigger our "EmptyTable" Component, with several options on what the user wants to create as default?

EDIT: I just saw the issue reference in your comment, i would highly recommend to just do an empty table then, and reusing the empty table naming schema from the plus button

src/Shared/ARCtrl.Helper.fs
```diff
@@ -155,6 +168,19 @@ module ARCtrlHelper =
             | ArcFiles.Run _ -> true
             | _ -> false
 
+        member this.EnsureDefaultAnnotationTable() =
```
Freymaurer: This should not be a "defaultAnnotationTable" to be ensured, i am not really sure if this even deserves its own function. As this is simply ArcFiles.ArcTables().InitTable(createNewTableName xxx).

Just fully merge it into `createDefaultArcFile`, for which it makes sense again. Because then we say "if you create a new file of this type it comes with an empty table."

Freymaurer (changes requested):

src/Shared/ARCtrl.Helper.fs
Etschbeijer: Fixed.
Made it a single source of truth.

src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
```diff
@@ -133,6 +144,38 @@ module ArcLoadExtensions =
         return repairedAny
     }
 
+    let private addRecoveredEmptyAnnotationTables (arc: ARC) (contracts: Contract[]) =
```
Freymaurer: Are we sure we want this? This is highly specific and can lead to unexpected stuff. 

This si not on the level of "initialize new files always with empty table". This actually redefined full read behavior. 

In my opinion this must be discussed first and split in a new issue if desired

tests/Electron.Core/ArcAddExtensions.test.fs
```diff
@@ -15,12 +15,26 @@ let private expectSome (value: 'T option) (message: string) : 'T =
...
+    | Ok arc -> return arc
+    | Error errors -> return failwith (PathHelpers.formatContractErrors errors)
+}
+
 let private testWindow = TestHelpers.testWindow
 
 let private withTempArc =
     TestHelpers.withTempArcWith "swate-add-arc-file-" "AddArcFileArc"
 
+let private createDefaultArcFile fileType identifier =
```
Freymaurer: > - Anything added dilutes everything else

Please do not add wrappers if the yare not absolutly required! Every single wrapper forces me during pr review to relearn code. If we always use the same function calls with different arguments it is soooo much faster for me to see what is happening. If there is a single parameter change, do not add wrappers as well.

src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
```diff
@@ -48,26 +51,34 @@ module ArcLoadExtensions =
         CreateContracts: string -> Contract[]
     }
 
+    let private createDefaultArcFileContracts (fileType: ArcFilesDiscriminate) (identifier: string) =
```
Freymaurer, line 52: this is code duplocation, it exists in `src/Shared/ARCtrl.Helper.fs`

Freymaurer (changes requested):

tests/Electron.Core/ArcAddExtensions.test.fs
Etschbeijer: fixed

src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
Etschbeijer: fixed

src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
Etschbeijer: fixed

tests/Electron.Core/ArcAddExtensions.test.fs
Etschbeijer: Fixed in commit: 53869d5101307d3c80b9c3a1505b70f42133698d

src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
Etschbeijer: fixed in commit: 53869d5101307d3c80b9c3a1505b70f42133698d

src/Electron/src/Main/ARCtrlExtensions/LoadAsync.fs
Etschbeijer: Fixed in commit: 53869d5101307d3c80b9c3a1505b70f42133698d

Freymaurer (approved):

PR #1198, opened 2026-06-23
Add handling of dirty notes when leaving them

Etschbeijer:
* When leaving a dirty note, a modal is called that enables the user to proceed without saving, cancel the leaving or save the note and leave


src/Components/src/Composite/Notes/Editor/Conversion.fs
```diff
@@ -186,3 +186,40 @@ module NoteConversion =
             $"{header}\n"
         else
             $"{header}\n\n{body}\n"
+
+    let private createPayloadWithDate (target: NotesTarget) (relativePath: string) dateCreated (draft: NotesDraft) = {
```
Freymaurer: This should be a static member on the type. This allows us to set some of the input params as option and provide sensible defaults, such as "no date is given, we use now"

src/Components/src/Composite/Notes/Editor/Conversion.fs
```diff
@@ -186,3 +186,40 @@ module NoteConversion =
...
+            Target = target
+        }
+        Title = draft.Title.Trim()
+        DateCreated = dateCreated
+        Tags = draft.Tags |> Seq.toList
+    }
+
+    let private tryCreatePayload target resolveRelativePath unsafePathMessage draft =
+        if Validation.isRequiredDataValid draft |> not then
+            Error "Please enter a Title and a Date Created value before submitting."
```
Freymaurer: The first 3 errors should be handled before this is even called. As this function can just specifiy, that these MUST NOT be options.

In the ui we do not want users to click a button if it will oviously fail. Just disable the button if the values do not exist and in the onclick check if they exist and if not, just do nothing.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -34,6 +34,8 @@ type ArcVault(window: BrowserWindow) =
     //I am not sure if performance of calculating changes or precision should be more important. Lets see in the future. Maybe you can add this comment as /// comment on this member?
     /// Dirty marker for unsaved in-memory ARC mutations.
     member val hasUnsavedArcChanges: bool = false with get, private set
+    /// Dirty marker for unsaved renderer-owned markdown note edits.
+    member val hasUnsavedNoteChanges: bool = false with get, set
```
Freymaurer: No! Track unsaved notes per file, not with the full ARC. This is something that should be done for all single file elements. Think vs-code, you MUST save each file and will be asked FOR the file, when you try to close it with unsaved changes.

Compare with the ArcFiles logic to a .xlsx file. Multiple files that build on top of each other. it makes sense to only save it together. 

The unsavedChanges feature might be something that we will have to revisit in the future to move it completly outside of the ArcVault

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -542,6 +542,19 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
             with e ->
                 return Error e
         }
+    setHasUnsavedNoteChanges =
```
Freymaurer: as written above. do not store this in ArcVault.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -605,7 +607,7 @@ type ArcVaults() =
 
         window.onClose (fun closeEvent ->
             if not vault.isCloseApproved then
-                if vault.hasUnsavedArcChanges then
+                if vault.hasUnsavedArcChanges || vault.hasUnsavedNoteChanges then
```
Freymaurer: the function, which sets `vault.isCloseApproved`, checks this against the renderer state. It should, in addition, check if a notes page is open with changes, and if that is the case return a specialized state and not approve!

Freymaurer (changes requested):

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: f6df69431cebf83cf98c592b2033495807f98d41

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: f6df69431cebf83cf98c592b2033495807f98d41

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: a668f52b3f002564ae67f902bc85e476b4b5f073

src/Components/src/Composite/Notes/Editor/Conversion.fs
Etschbeijer: Fixed with commit: 7232d8409495b3fe457c39d641f765682cdb110d

src/Components/src/Composite/Notes/Editor/Conversion.fs
Etschbeijer: Fixed with commit: 05b326b5f094072c866b6bc4f7fac5495459f467

Issue #1157, opened 2026-06-15
[Feature Request]: add link to ARC in DataHUB


Brilator:
### Host

Electron Swate App

Brilator:
Ok, I just found the `open origin repository`. 

This could

- [ ] be more prominent
- [ ] say `open ARC in DataHUB`
- [ ] use the globe icon as in ARCitect

Freymaurer:
We could add this option into the ArcSummary component (on top of file tree), but it is rather low priority. The icon is used by us for any external references, but the renaming makes sense. Could you open a PR for this?

Freymaurer:
🚀

PR #1201, opened 2026-06-23
change label to open ARC in datahub

Brilator:
#1157


Brilator:
What? I just changed two lines...

PR #1208, opened 2026-06-24
Feature/table editor

caroott:
This PR includes the current state of the ProvenanceGrouping (WiP) and a composite component that organizes draggable items into expandable folders within a drag-and-drop shelf.

### What it does

Displays a horizontal row of folders. Clicking a folder expands it, revealing its items in a drop shelf below. Each item can be dragged to external drop targets, carrying structured metadata about both the item and its parent folder. External items can also be dropped into an expanded folder's shelf.

### Key behaviors

- **Single-folder expansion**: only one folder can be open at a time. Clicking a different folder collapses the current one.
- **Cascading colors**: items inherit their folder's color unless they specify their own override. Dragged items carry an `EffectiveColor` that resolves this fallback.
- **External drop acceptance**: when a folder is expanded and the consumer provides a handler, external drag sources can drop items into that folder's shelf. The handler receives the target folder, the source drag ID, and the source drag data.
- **Disabled items**: individual items can be marked as disabled. They render as visual placeholders but are not draggable.
- **Drag overlay**: while dragging, the source item becomes transparent and a styled copy follows the pointer with a drop shadow.
- **Controlled or uncontrolled expansion**: the expanded folder can be managed externally via `expandedFolderIds`/`onExpandedFolderIdsChange`, or the component manages it internally via `defaultExpandedFolderIds`.
- **Tooltips and badges**: items support optional tooltips (shown on hover) and badges (small count labels).

### Usage

| Prop | Type | Required | Description |
|---|---|---|---|
| `folders` | `FolderedDraggableFolder<'payload> list` | ✅ | The list of folders, each containing items |
| `dragId` | `folder -> item -> string` | ✅ | Opaque stable ID for drag-and-drop identity |
| `expandedFolderIds` | `Set<string>` | — | Controlled: set of expanded folder IDs |
| `defaultExpandedFolderIds` | `Set<string>` | — | Uncontrolled: initial expanded folder IDs |
| `onExpandedFolderIdsChange` | `Set<string> -> unit` | — | Called when expansion changes |
| `renderItemContent` | `render -> ReactElement` | — | Custom render function for item content |
| `shelfDropId` | `string` | — | Custom droppable ID for the shelf zone |
| `tryCreateItemFromExternalDrop` | `drop -> item option` | — | Handler to convert external drag data into a new item |
| `onFoldersChange` | `folders list -> unit` | — | Called when items are added via external drop |
| `className` | `string` | — | Additional CSS classes on the root element |
| `debug` | `bool` | — | Adds `data-testid` attributes throughout |

### Trade-offs

- The component renders the **read-only state** (existing items). The consumer is responsible for mutation (add/remove items) based on drag events.
- Internal item-to-item reordering within a folder is **not supported**. Only drag-in (external → shelf) and drag-out (shelf → external) are modeled.
- The shelf's drop zone is only active when a folder is expanded AND both `tryCreateItemFromExternalDrop` and `onFoldersChange` are provided.

### DIFF
--- a/src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
+++ b/src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
@@ -88,7 +88,7 @@ module private FolderedDraggableListHelper =
         else
             "swt:btn-outline swt:bg-base-100 swt:cursor-grab swt:shadow-sm active:swt:cursor-grabbing"
         if isDragging then
-            "swt:absolute swt:pointer-events-none swt:opacity-0"
+            "swt:pointer-events-none swt:opacity-0"
     ]
 
 [<Erase; Mangle(false)>]
@@ -159,12 +159,15 @@ type FolderedDraggableList =
                 | _ -> ()
             ]
             prop.ariaLabel $"Drag {item.Label}"
+            if draggable.isDragging then
+                prop.custom ("aria-hidden", true)
+                prop.tabIndex -1
             match item.Tooltip with
             | Some tooltip when tooltip <> "" ->
                 prop.custom ("data-tip", tooltip)
                 prop.title tooltip
             | _ -> ()
-            if defaultArg debug false then
+            if defaultArg debug false && not draggable.isDragging then
                 prop.testId $"foldered-draggable-item-{item.Id}"
             prop.children [ renderItemContent renderProps ]
         ]
@@ -309,8 +312,11 @@ type FolderedDraggableList =
 
         DndKit.useDndMonitor (
             {|
+                onDragCancel = fun (_: DndKit.IDndKitEvent) -> setActiveDrag None
                 onDragEnd =
                     fun (event: DndKit.IDndKitEvent) ->
+                        setActiveDrag None
+
                         match expandedFolder, tryCreateItemFromExternalDrop, onFoldersChange with
                         | Some targetFolder, Some tryCreateItemFromExternalDrop, Some onFoldersChange when
                             FolderedDraggableListHelper.isShelfDrop shelfDropId event
@@ -400,6 +406,7 @@ type FolderedDraggableList =
                     ]
                 ]
                 DndKit.DragOverlay(
+                    dropAnimation = {| duration = 0; easing = "linear" |},
                     children =
                         match activeDrag with
                         | Some render ->
--- a/src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
+++ b/src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
@@ -184,6 +184,13 @@ type FolderedDraggableList =
                 else
                     "swt:border-base-300 swt:bg-base-100 hover:swt:border-primary/60 hover:swt:bg-base-200"
             ]
+            match folder.Color with
+            | Some color when color <> "" ->
+                prop.custom ("data-foldered-folder-color", color)
+
+                if not isExpanded then
+                    prop.style [ style.borderColor color ]
+            | _ -> ()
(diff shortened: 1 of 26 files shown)


Freymaurer:
Will not fully review this at this point in time, but here are some pointers:

- try to use more js friendly types: POJO, anonymous records wherever you might run into performance concerns.
- try to reduce usage of "useEffect", there is a really good article linked in design docs on epic branch

src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
```diff
@@ -0,0 +1,426 @@
...
+
+[<Erase; Mangle(false)>]
+type FolderedDraggableList =
+
+    [<ReactComponent>]
+    static member private Item<'payload>
+        (
+            folder: FolderedDraggableFolder<'payload>,
+            item: FolderedDraggableItem<'payload>,
+            dragId: FolderedDraggableFolder<'payload> -> FolderedDraggableItem<'payload> -> string,
```
Etschbeijer, line 142: Is dragId the startId or targetId?

src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
```diff
@@ -0,0 +1,426 @@
...
+            ]
+        ]
+
+    [<ReactComponent>]
+    static member FolderedDraggableList<'payload>
+        (
+            folders: FolderedDraggableFolder<'payload> list,
+            dragId: FolderedDraggableFolder<'payload> -> FolderedDraggableItem<'payload> -> string,
+            ?expandedFolderIds: Set<string>,
+            ?defaultExpandedFolderIds: Set<string>,
```
Etschbeijer, line 391: What are defaultExpandedFolderIds?

src/Components/src/Page/ProvenanceGrouping/Controls.fs
```diff
@@ -30,6 +31,69 @@ module private SideLabels =
         | ProvenanceSide.Input -> "input"
         | ProvenanceSide.Output -> "output"
 
+module private ColorPicker =
+
+    let fallbackColor =
+        State.PropertyColors.palette |> Array.tryHead |> Option.defaultValue "#2563eb"
```
Etschbeijer, line 37: Why fallbackColour and not defaultColour? It is always the same colour or not?

src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
Etschbeijer, line 1: Maybe we can rename the whole file. A file with the same name as the folder is not a good practice or is it?
Main might be fitting.

Etschbeijer (changes requested):

src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
caroott: image

src/Components/src/Page/ProvenanceGrouping/Controls.fs
caroott: image

src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
caroott: image

src/Components/src/Composite/FolderedDraggableList/FolderedDraggableList.fs
caroott: The dragId is the source/start ID. It is passed to useDraggable({ id = dragId folder item, ... }). When the item is dragged, event.active.id equals this string.

PR #1215, opened 2026-06-25
Fix git upload problems

caroott:
Two improvements to Git push/pull reliability: cancelable LFS uploads with idle-based timeouts, and a fix for LFS authentication failures caused by `http.extraHeader` appendede to LFS transfer commands.

### Cancel button for LFS uploads

- Added a **Cancel** button in the Git sidebar, shown during LFS upload (`Uploading Git LFS objects` phase).
- Replaced **flat timeouts** on spawned Git processes with **idle timeouts**. The timer resets on every stdout/stderr event, so  transfers with progress output won't time out prematurely. Processes that go silent for the timeout duration are still killed.
  - Push/Pull don't have timeouts

### Fix: LFS auth header stripping

- Git LFS uses its own HTTP authentication separate from regular Git HTTP auth. When `http.extraHeader` (set by the credential helper) was passed to `git lfs push/pull/smudge`, some LFS servers rejected the conflicting header.
- Added a filter (`lfsTransferConfigArgs`) that strips `http.extraheader` and `http.<url>.extraheader` config args before spawning LFS transfer commands.
- The filter is applied at all 4 LFS transfer call sites: `pullAll`, `fetchRefetchForPath`, `downloadObjectFromListing`, and `uploadObjects`.

### DIFF
--- a/src/Components/src/Page/GitSidebar/GitSidebar.fs
+++ b/src/Components/src/Page/GitSidebar/GitSidebar.fs
@@ -262,9 +262,27 @@ type GitSidebar =
             ?warningNotice: string,
             ?busyTestId: string,
             ?errorTestId: string,
-            ?warningTestId: string
+            ?warningTestId: string,
+            ?onCancelOperation: unit -> unit
         ) =
         let runStatus = defaultArg runStatus GitSidebarRunStatus.Idle
+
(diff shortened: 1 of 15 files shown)


Etschbeijer (approved):
Lgtm

PR #1217, opened 2026-06-29
Follow-up LFS reliability fix

caroott:
Follow-up LFS reliability fix for #1215 : prevent Git workflows from invoking LFS implicitly outside Swate’s authenticated transfer path.

## Changes

- Skip Git LFS pre-push hooks during `git push --dry-run` planning with `GIT_LFS_SKIP_PUSH=1`.
  - This prevents dry-run push planning from attempting an unauthenticated LFS upload before Swate reaches its explicit LFS upload step.

- Skip implicit LFS smudge during worktree-changing Git operations with `GIT_LFS_SKIP_SMUDGE=1`.
  - Applied to branch creation, branch checkout, and discard/restore paths.
  - Keeps LFS files as pointer files unless Swate explicitly downloads them through the authenticated LFS flow.

### DIFF
--- a/src/Electron/src/Main/Git/GitLfsService.fs
+++ b/src/Electron/src/Main/Git/GitLfsService.fs
@@ -965,7 +965,9 @@ let planOutboundPush
         let! dryRunResult =
             runSimpleGitRaw
                 (fun currentGit ->
-                    currentGit.raw [|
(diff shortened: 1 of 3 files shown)


Etschbeijer (approved):
Lgtm

PR #1231, opened 2026-07-03
Make parent of branch sticky in filetree

Etschbeijer:
* Track each rendered item’s tree depth
* Use that depth to calculate a stacked sticky top offset for expanded parent branches
* Keep expanded parent branches sticky so their collapse controls remain visible while scrolling

### DIFF
--- a/src/Components/src/Page/FileExplorer/FileExplorer.fs
+++ b/src/Components/src/Page/FileExplorer/FileExplorer.fs
@@ -158,6 +158,17 @@ type FileExplorer =
         let model, dispatch = React.useReducer (reducer, initialModel)
         let containerRef = React.useElementRef ()
 
+        let stickyRowHeights, setStickyRowHeights =
+            React.useStateWithUpdater (Map.empty<string, int>)
+
+        let handleStickyRowHeightChange (itemId: string) (height: int) =
+            setStickyRowHeights (fun heights ->
+                match heights |> Map.tryFind itemId with
+                | Some currentHeight when currentHeight = height -> heights
+                | _ -> heights |> Map.add itemId height
+            )
+            |> ignore
+
         let onDirectoryExpansionChange =
             onDirectoryExpansionChange
             |> Option.orElse onExpansionChange
@@ -271,7 +282,7 @@ type FileExplorer =
 
         let selectedPathIds = model.SelectedPath |> List.map _.Id |> Set.ofList
 
-        let rec renderItem depth item =
+        let rec renderItem depth stickyTopOffset item =
             let isSelected = model.SelectedId = Some item.Id
             let isInSelectedPath = selectedPathIds.Contains item.Id
             let isHighlighted = isSelected || isInSelectedPath
@@ -303,10 +314,17 @@ type FileExplorer =
                     if isExpanded then
                         match item.Children with
                         | Some children ->
+                            let childStickyTopOffset =
+                                if delegateHorizontalScrollToParent then
+                                    stickyTopOffset
+                                    + (stickyRowHeights |> Map.tryFind item.Id |> Option.defaultValue 0)
+                                else
+                                    stickyTopOffset
+
                             Some(
                                 Html.ul [
                                     prop.className "swt:ml-4"
-                                    prop.children (children |> List.map (renderItem (depth + 1)))
+                                    prop.children (children |> List.map (renderItem (depth + 1) childStickyTopOffset))
                                 ]
                             )
                         | None -> None
@@ -316,6 +334,18 @@ type FileExplorer =
                 let stickyDepth =
                     if delegateHorizontalScrollToParent then
                         Some depth
+                    else
+                        None
+
+                let rowStickyTopOffset =
+                    if delegateHorizontalScrollToParent then
+                        Some stickyTopOffset
+                    else
+                        None
+
+                let onStickyRowHeightChange =
+                    if delegateHorizontalScrollToParent then
+                        Some handleStickyRowHeightChange
                     else
                         None
 
@@ -340,6 +370,8 @@ type FileExplorer =
                     canDeleteItem = canDeleteItem,
                     ?statusAction = statusAction,
                     ?stickyDepth = stickyDepth,
+                    ?stickyTopOffset = rowStickyTopOffset,
+                    ?onStickyRowHeightChange = onStickyRowHeightChange,
                     ?children = childrenTree
                 )
             else
@@ -366,7 +398,7 @@ type FileExplorer =
                         Html.ul [
                             prop.testId "file-explorer-container"
                             prop.className listClassName
-                            prop.children (model.Items |> List.map (renderItem 0))
+                            prop.children (model.Items |> List.map (renderItem 0 0))
                         ]
                     ]
                 ]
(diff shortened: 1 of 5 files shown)


Etschbeijer:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

No issue associated with this PR.

## Type of change

- [ X ] New feature (non-breaking change which adds functionality)

- [ X ] Electron - Renderer

# How Has This Been Tested?

- [x] Manual test: I opened a filetree with a big dataset, opened that and scrolled down.
- [x] I added a storybook test that does basically the same.

# Checklist:

- [ X ] My code follows the style guidelines of this project
- [ X ] I have performed a self-review of my code
- [ X ] I have commented my code, particularly in hard-to-understand areas
- [ X ] I have made corresponding changes to the documentation
- [ X ] I have added tests that prove my fix is effective or that my feature works

> The Zen of GitHub.

- Practicality beats purity
- Speak like a human
- Encourage flow
- Favor focus over features
- Design for failure
- Keep it logically awesome

Etschbeijer:
grafik

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -285,6 +286,12 @@ type FileExplorerItem =
         let canDeleteFromDirectory = canDeleteItem item && onDeleteItem.IsSome
         let hasStatusControl = Helper.isLfs item || statusAction.IsSome
 
+        let stickyRowClasses =
+            if stickyTopOffset.IsSome then
```
Freymaurer: make the design `swt:bg-base-100 swt:border-b swt:border-base-content/10 swt:shadow-sm` reliant on isExpanded/isOpen/isNotCollapsed (no idea what it is called), move "swt:sticky" to style `prop.style [style.position.sticky]` to make it have priority and non reliant on isOpen.

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -296,9 +303,21 @@ type FileExplorerItem =
...
                     rowHighlightClass
                 ]
-                prop.style [ style.display.flex; style.width (length.percent 100) ]
+                prop.style [
+                    style.display.flex
+                    style.width (length.percent 100)
+                    yield!
+                        stickyTopOffset
+                        |> Option.map (fun offset -> [
+                            style.position.sticky
```
Freymaurer, line 338: ahh it is already here!

src/Components/src/Page/FileExplorer/FileExplorer.fs
```diff
@@ -330,6 +340,7 @@ type FileExplorer =
                     ?onDeleteItem = onDeleteItem,
                     canDeleteItem = canDeleteItem,
                     ?statusAction = statusAction,
+                    ?stickyTopOffset = stickyTopOffset,
```
Freymaurer: if it can only be open and sticky at the same time, it might be possible to unify those two. Maybe add `depth` to the internal data model. Then we can calculate it from the row component. 

Make sure this change makes sense, i am not sure how well this can be done. if it becomes to abstract, don't change it

Freymaurer (changes requested):

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -296,9 +303,21 @@ type FileExplorerItem =
...
+                            style.top offset
+                            style.zIndex (100 - (offset / 32))
```
Freymaurer: stickyParentRowHeightPx

Freymaurer: and add a comment for this calculation that it might run into layering issues with high numbers of depth? At least i assume this will happen

Freymaurer (changes requested):

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
Etschbeijer: Fixed in commit: 8377f284b40fec6845ac334b98951fba53cf2a13

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
Etschbeijer: Fixed in commit: fc15b88f3409aa3f5c562f7038a17e1e28da41b4

src/Components/src/Page/FileExplorer/FileExplorer.fs
Etschbeijer: fc15b88f3409aa3f5c562f7038a17e1e28da41b4

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
Etschbeijer: Updated comment in commit: a4d2798051a3935fa516749a1cbe251ebf9b3f86

Simplified the calculation.

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -296,9 +301,24 @@ type FileExplorerItem =
                 prop.custom ("data-file-item-id", item.Id)
                 prop.className [
                     "swt:group swt:w-full swt:px-2 swt:py-1 swt:cursor-default"
+                    if isExpanded then
```
caroott: the background/border/shadow is gated purely on isExpanded.
But sticky positioning is gated on effectiveStickyDepth (i.e. isExpanded && stickyDepth.IsSome, which is only ever Some when delegateHorizontalScrollToParent = true). The other consumers of this component do not set delegateHorizontalScrollToParent, so their expanded directory rows now get a solid background, bottom border, and drop shadow with no sticky behavior to justify it.
@Freymaurer asked to tie the design to isExpanded instead of stickyTopOffset.IsSome, but almost certainly meant within the sticky feature.

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
```diff
@@ -285,6 +286,10 @@ type FileExplorerItem =
         let canDeleteFromDirectory = canDeleteItem item && onDeleteItem.IsSome
         let hasStatusControl = Helper.isLfs item || statusAction.IsSome
 
+        let effectiveStickyDepth = if isExpanded then stickyDepth else None
+
+        let stickyParentRowHeightPx = 32
```
caroott: stickyParentRowHeightPx is set to 32, but the row is swt:px-2 swt:py-1 around content whose tallest element is the h-5 (20px) toggle button, plus the new 1px border-b, which is closer to ~29px than 32. Nothing enforces the coupling, so if the row's padding/content ever changes, the stack silently misaligns (gaps show scrolled content between pinned rows, or rows overlap). The Storybook test only asserts toggles stay within the viewport bounds. It won't catch stacking gaps/overlap, so this can regress unnoticed. Consider deriving the offset from a measured height, or at minimum a comment tying the constant to the row's py-1/h-5 so a future edit knows to update it.

caroott (changes requested):

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
Etschbeijer: Fixed in commit: da40fce16b2b563186ede229019fab8f7bb48530

src/Components/src/Page/FileExplorer/FileExplorerItem.fs
Etschbeijer: Fixed in commit: 5fa2a63e887009dcc046179c1a073017b2aa5c74

PR #1255, opened 2026-07-14
Fix Property assignment and overwrite

caroott:
## Property inheritance and origin-aware overwrites in the provenance editor

This PR fixes two related problems in the provenance grouping editor: property values
not staying in sync across connected endpoints, and same-named properties from
different source tables being treated as one property.

### Fix connection annotation inheritance

Inheritance is again bidirectional: a value on any endpoint spreads through the layer's
connections to every transitively connected endpoint, no matter which side it lives on.
Removing a connection retracts exactly the values that arrived through it, and values
carried in from upstream layers are left untouched. An endpoint never inherits its own
values back.

### Same header, different source = different property

"Temperature" from the Growth table and "Temperature" from the Cultivation table used
to collapse into a single property. Grouping by one would mix in the other's values,
and overwriting one could silently touch both.

A property's identity is now its header *plus* the source it originated in. Each origin
gets its own rail row, its own color, its own drag identity, and its own connector
lines. So two same-named properties stay visually and behaviorally separate throughout
the editor.

### Overwrites respect the property's origin

Dropping a replacement value onto a group now only targets existing assignments of that
exact property:

- A same-named property from another source can neither join nor block the overwrite.
- When one value is represented by several writeback occurrences (e.g. through
  inheritance), all of them are updated together.
- A property that originated upstream can only *replace* existing assignments
  downstream. It can't be attached to entities that never had it. Trying to do so
  shows a clear error instead of quietly creating a new assignment.
- If the targeted entities don't share a single existing value, the drop is rejected,
  because there is no one value that can be replaced safely.

### DIFF
--- a/src/Components/src/Page/ProvenanceGrouping/ConnectorMeasurement.fs
+++ b/src/Components/src/Page/ProvenanceGrouping/ConnectorMeasurement.fs
@@ -328,10 +328,10 @@ module ConnectorHandles =
         ParentGroupId = Some groupId
     }
 
-    let propertyHeader side header : ConnectionHandleRef = {
+    let propertyHeader side property : ConnectionHandleRef = {
         Kind = ConnectionHandleKind.PropertyHeader
         Side = side
-        Id = DragDrop.propertyHeaderIdentity header
+        Id = DragDrop.propertyKeyIdentity property
         ParentGroupId = None
     }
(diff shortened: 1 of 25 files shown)


Etschbeijer (approved):
Lgtm

Issue #1277, opened 2026-08-03
[Bug]: editing one overwrites all isa*xlsx files


Brilator:
### What happened?

https://github.com/user-attachments/assets/434e85b4-63e9-497e-a5b1-cc0fa962b40c

SwateApp

related
- https://github.com/nfdi4plants/ARCitect/issues/129
- https://github.com/nfdi4plants/ARCitect/issues/171

### Host

Applies to All

Etschbeijer:
Error replication:
* Create a new ARC
* Create two assays
* Adapt one assay in excel
* Adapt the same assay in SwateApp and store it
* The second Assay and investigation file will also be displayed as changed

PR #1287, opened 2026-08-04
Bugfix: Single, external xlsx files do only change one file

Etschbeijer:
Goal: Make sure that only the adaped XLSX files are loaded into git and not unaltered XLSX files

* The reloads of the file watch can lose the static hash baseline during an ARC merge. As a result, unchanged entities are treated as modified files and multipleXLSX files are overwritten when saving. 
* All of them appear in the GIT changes then.
* The changes preserve the disk hash baseline across watcher merges but also keep local changes detectable


Freymaurer (approved):
Lgtm

Etschbeijer:
Closes issue #1277

Etschbeijer:
Add change of this pr to next pr

PR #1295, opened 2026-08-05
Expand datamap update to table

Etschbeijer:
* Store File Picker and Data Annotator references with an explicit `./` ARC-root-relative path in both table and DataMap views
* Preserve data selectors after `#` when copying single or multiple cells from the normal table view

### DIFF
--- a/src/Shared/ARCtrl.Helper.fs
+++ b/src/Shared/ARCtrl.Helper.fs
@@ -227,7 +227,11 @@ module ARCtrlHelper =
 
         let withExplicitRelativePrefix (path: string) =
             let normalizedPath = path.Replace('\\', '/')
-            if normalizedPath.StartsWith("./") then normalizedPath else $"./{normalizedPath}"
+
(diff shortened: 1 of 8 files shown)


src/Shared/ARCtrl.Helper.fs
```diff
@@ -225,6 +225,14 @@ module ARCtrlHelper =
             | ArcFiles.DataMap(Some parentInfo, _) -> Some parentInfo
             | _ -> None
 
+        let withExplicitRelativePrefix (path: string) =
```
caroott, line 228: don't we have this normalize function several times by now?

caroott (approved):
looks ok, same remark as previous one

src/Shared/ARCtrl.Helper.fs
Etschbeijer: This one calls the normal path normalization and updates it in a way that the others do not.

PR #1299, opened 2026-08-06
Add Datamap create button

Etschbeijer:
* Add a Datamap create button to the file explorer view
* Increate version to 2.0.7

### DIFF
--- a/src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
+++ b/src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
@@ -125,8 +125,7 @@ let ArcFilePreviewTarget (arcFile: ArcFiles, startingActiveView: ActiveView opti
         promise {
             match! Helper.saveArcFile nextArcFile with
             | Ok() -> ()
-            | Error exn ->
-                errorModal.enqueue (ErrorModalRequest.create (exn.Message, title = "Could not save DataMap"))
+            | Error exn -> errorModal.enqueue (ErrorModalRequest.create (exn.Message, title = "Could not save DataMap"))
         }
         |> Promise.start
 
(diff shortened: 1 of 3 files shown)


Freymaurer (approved):

src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
```diff
@@ -118,10 +164,17 @@ let ArcFilePreviewTarget (arcFile: ArcFiles, startingActiveView: ActiveView opti
     let trailingNavbarElements =
         React.useCallback ((fun props -> TableNavbarActions(props, setArcFile)), [| box setArcFile |])
 
+    let widgetNavbarElements =
```
caroott, line 167: This uses a dependency that is a plain lambda rebuilt on every render, so it never compares equal. That invalidates the new `widgetNavbarElement` memo, which in turn invalidates the `navbar` memo in `ArcFileEditor.fs`.

Either wrap `saveArcFileImmediately` in a `useCallback` with real dependencies, or drop the memos so the cost is visible.

src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
```diff
@@ -54,6 +54,42 @@ let private TableNavbarActions (props: ArcFileEditorHeaderProps, setArcFile: Arc
...
+        | ArcFiles.Workflow workflow -> true, workflow.DataMap.IsNone
+        | _ -> false, false
+
+    let addDataMap _ =
+        if canAddDataMap then
+            let dataMap = Some(ARCtrl.DataMap.init ())
+
+            match props.arcFile with
+            | ArcFiles.Assay assay -> assay.DataMap <- dataMap
+            | ArcFiles.Study(study, _) -> study.DataMap <- dataMap
```
caroott, line 73: `addDataMap` mutates the arcFile held in state before copying it

`addDataMap` writes `assay.DataMap <- dataMap` directly onto `props.arcFile` and only then calls `ArcFiles.refreshRef`. The result is correct because the copy is taken after the write, but the previous state object is mutated too, so any code still holding it sees the change.

Prefer copying first and mutating the copy, the way `updateArcFile` in `ArcFileFooterTabs.fs` does it after PR #1303.

src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
Etschbeijer: I fixed this part already in the PR 1301.

PR #1301, opened 2026-08-06
Refactor of recent changes

Etschbeijer:
* Refactor add datamap button
* Refactor Actionbar associated changes
* Refactor filetree associated changes
* Refavtor filetree pagestate handling

### DIFF
--- a/src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
+++ b/src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
@@ -55,53 +55,12 @@ let private TableNavbarActions (props: ArcFileEditorHeaderProps, setArcFile: Arc
     | _ -> Html.none
 
 [<ReactComponent>]
-let private CreateDataMapNavbarAction (props: ArcFileEditorHeaderProps, saveArcFile: ArcFiles -> unit) =
-    let supportsDataMap, canAddDataMap =
-        match props.arcFile with
-        | ArcFiles.Assay assay -> true, assay.DataMap.IsNone
-        | ArcFiles.Study(study, _) -> true, study.DataMap.IsNone
-        | ArcFiles.Run run -> true, run.DataMap.IsNone
-        | ArcFiles.Workflow workflow -> true, workflow.DataMap.IsNone
-        | _ -> false, false
-
-    let addDataMap _ =
-        if canAddDataMap then
-            let dataMap = Some(ARCtrl.DataMap.init ())
-
-            match props.arcFile with
-            | ArcFiles.Assay assay -> assay.DataMap <- dataMap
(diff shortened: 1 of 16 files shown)


Freymaurer (approved):
lgtm

src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
```diff
@@ -165,21 +114,63 @@ let ArcFilePreviewTarget (arcFile: ArcFiles, startingActiveView: ActiveView opti
...
+            | ArcFiles.Assay assay -> button assay.DataMap (fun value -> assay.DataMap <- value)
+            | ArcFiles.Study(study, _) -> button study.DataMap (fun value -> study.DataMap <- value)
+            | ArcFiles.Run run -> button run.DataMap (fun value -> run.DataMap <- value)
+            | ArcFiles.Workflow workflow -> button workflow.DataMap (fun value -> workflow.DataMap <- value)
+            | _ -> Html.none
+
+    let editorKey =
+        arcFile.TryGetRelativePath()
+        |> Option.defaultValue (string arcFile.RelatedArcFilesDiscriminate),
+        startingActiveView |> Option.map _.ViewIndex
```
caroott, line 154: `ArcFileEditor` silently ignores `startingActiveView` changes now

The effect that reset `activeView` when `startingActiveView` changed was removed from the shared component. Resetting now depends on a `prop.key` remount that lives in one specific caller.

I checked and `ArcFilePreviewTarget` is currently the only consumer besides the story, so nothing is broken today. But the public `?startingActiveView` parameter now only applies on mount and there is no doc comment saying that. The next consumer of this component will pass it, see nothing happen, and have to go read the Electron renderer to find out why.

Please either keep the reset inside the component or add an XML doc note on the parameter pointing at the remount contract.

src/Electron/src/Renderer/Components/MainContent/ArcFilePreviewTarget.fs
```diff
@@ -165,21 +114,63 @@ let ArcFilePreviewTarget (arcFile: ArcFiles, startingActiveView: ActiveView opti
         React.useCallback ((fun props -> TableNavbarActions(props, setArcFile)), [| box setArcFile |])
 
     let widgetNavbarElements =
-        React.useCallback (
```
caroott, line 168: `widgetNavbarElements` was de-memoized, which kills the `navbar` memo

`widgetNavbarElements` is now a bare `fun props -> ...` rebuilt on every render, and it is a dependency of the `navbar` memo. `headerProps` is properly stable, so this is the only dependency that always changes. The navbar, and therefore `content`, is rebuilt on every render.

Either restore a `useCallback` here or remove the memos in `ArcFileEditor`.

Issue #1300, opened 2026-08-06
[Bug]: Could not update ARC in memory


Brilator:
### What happened?

https://github.com/user-attachments/assets/3f8680af-28bf-4b96-be9a-1065f030052e

- Seems to be specific to the training templates. 
- Has some downstream issues, when trying to work with that annotation table.

Brilator:
probably related #1298

Etschbeijer:
I can reproduce this error.
When no table is available but you select append, we run in a brief memory crash, where the new table must be generated.
We should not allow appending when not table is available.
Add should be the only choice then.

Issue #1298, opened 2026-08-05
[Bug]: (minor) deleted tables pop up again


Brilator:
### What happened?

deleted tables pop up again when creating a new one 

https://github.com/user-attachments/assets/15fd3a64-e061-4b0e-b825-b15fadaddf0c

### Host

Applies to All

PR #1303, opened 2026-08-06
Fix table handling

Etschbeijer:
* Prevent the reappearing of a deleted table when a new one is added
* Prevent error message when a template is appended to an empty table

### DIFF
--- a/src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
+++ b/src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
@@ -22,15 +22,11 @@ module private ArcFileFooterTabsHelper =
 
     let tryParseTableDragId (i: string) =
         if i.StartsWith TableDragIdPrefix then
-            let indexStr = i.Substring(TableDragIdPrefix.Length)
-
-            match System.Int32.TryParse indexStr with
+            match System.Int32.TryParse(i.Substring(TableDragIdPrefix.Length)) with
             | true, index -> Some index
             | _ -> None
         else
             None
-
-    let resolveDropTargetTableIndex (targetId: string) = tryParseTableDragId targetId
 
     let tryGetDndEventId (eventNode: obj) =
         if isNull eventNode then
@@ -57,10 +53,7 @@ module private ArcFileFooterTabsHelper =
         match i with
         | "metadata" -> Some ActiveView.Metadata
         | "datamap" -> Some ActiveView.DataMap
-        | _ ->
-            match tryParseTableDragId i with
-            | Some index -> Some(ActiveView.Table index)
-            | None -> None
+        | _ -> tryParseTableDragId i |> Option.map ActiveView.Table
 
     module ContextMenu =
 
@@ -322,19 +315,14 @@ type ArcFileFooterTabs =
             React.useState (None: int option)
 
         let metadataTabLabel =
-            React.useMemo (
-                (fun () ->
-                    match arcFile with
-                    | ArcFiles.Assay _ -> "Assay"
-                    | ArcFiles.Study _ -> "Study"
-                    | ArcFiles.Investigation _ -> "Investigation"
-                    | ArcFiles.Run _ -> "Run"
-                    | ArcFiles.Workflow _ -> "Workflow"
-                    | ArcFiles.Template _ -> "Template"
-                    | ArcFiles.DataMap _ -> "Datamap"
-                ),
-                [| box arcFile |]
-            )
+            match arcFile with
+            | ArcFiles.Assay _ -> "Assay"
+            | ArcFiles.Study _ -> "Study"
+            | ArcFiles.Investigation _ -> "Investigation"
+            | ArcFiles.Run _ -> "Run"
+            | ArcFiles.Workflow _ -> "Workflow"
+            | ArcFiles.Template _ -> "Template"
+            | ArcFiles.DataMap _ -> "Datamap"
 
         let setEditorMode =
             React.useCallback (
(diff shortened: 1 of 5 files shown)


Etschbeijer:
Closes issue #1298
Closes issue #1300

Freymaurer (approved):

src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
```diff
@@ -220,7 +215,12 @@ type ArcFileFooterTabs =
     [<ReactMemoComponent(AreEqualFn.FsEqualsButFunctions)>]
     static member private PlusBtn(onClick) =
```
caroott, line 216: `PlusBtn` never re-renders, which could be the root cause

`[<ReactMemoComponent(AreEqualFn.FsEqualsButFunctions)>]` compiles to `memo(PlusBtn, equalsButFunctions)`, and the generated comparator skips every function valued prop:

    result = ((typeof xValue === 'function') ? true : equals(xValue, y[key]));

`PlusBtn`'s props are `{ onClick }`, which is nothing but a function, so the comparator always returns true.

So `PlusBtn` never re-renders after mount and keeps the first render's `addNewTable` closure forever. That reproduces the reported bug exactly, and it explains why `RenameThenDeleteAndAddTable` was needed as a second story. Rename mutates the original object, delete then operates on the copy, and the frozen plus handler still points at the original, which still contains the deleted table.

The `arcFileRef` fix makes the stale closure read fresh state, so the symptom is gone. But `PlusBtn` is still frozen, so `canAddTable`, `setArcFile` and `closeEditorMode` inside `addNewTable` are still first render captures. They happen to be stable today.

Maybe drop the memo from `PlusBtn`, it is a static icon button with one handler and gains nothing from memoizing.

src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
```diff
@@ -293,34 +293,31 @@ type ArcFileFooterTabs =
...
+    [<ReactComponent>]
     static member Main
         (arcFile: ArcFiles, activeView: ActiveView, setActiveView: ActiveView -> unit, setArcFile: ArcFiles -> unit)
         =
         let tables = arcFile.ArcTables()
         let canAddTable = arcFile.CanCreateTables()
         let canRenderDataMap = arcFile.CanRenderDataMapView()
         let tabsRef = React.useElementRef ()
+        let arcFileRef = React.useRef arcFile
+        arcFileRef.current <- arcFile
```
caroott, line 305: `arcFileRef.current <- arcFile` runs in the render body**

React documents this as unsupported. It breaks StrictMode double rendering and concurrent rendering, because the render pass is supposed to be side effect free.

It also cannot be moved into an effect, since the whole point is to be fresh before effects run. If the memo issue is addressed, this ref can go away and the handlers can just close over the current `arcFile` again.

src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
```diff
@@ -357,32 +354,33 @@ type ArcFileFooterTabs =
...
+                            let nextName = createNewTableName nextTables.Tables
+                            nextTables.AddTable(ArcTable.init nextName)
+                            nextTables.TableCount - 1
+                        )
 
-                    setArcFile (ArcFiles.refreshRef arcFile)
-                    setActiveView (ActiveView.Table(tables.TableCount - 1))
+                    setActiveView (ActiveView.Table nextTableIndex)
 
         let deleteTable (tableIndex: int) =
```
caroott, line 367: two "delete table" implementations now

`deleteSelectedTable` still does `arcFile.ArcTables().RemoveTableAt tableIndex` followed by `refreshRef`, which is the pattern replaced by `updateArcFile` in the footer.

It is safe in practice, because its `props.arcFile` arrives fresh from the non memoized `TableNavbarActions`. But it still mutates the object held in React state before copying it, and it always jumps to `Metadata` instead of shifting the active index the way `deleteTable` does.

src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
```diff
@@ -404,10 +402,7 @@ type ArcFileFooterTabs =
 
 
         let tableIds =
-            React.useMemo (
-                (fun () -> tables |> Seq.mapi (fun index _ -> mkTableDragId index) |> ResizeArray),
-                [| box tables.TableCount; box tableNamesKey |]
-            )
+            tables |> Seq.mapi (fun index _ -> mkTableDragId index) |> ResizeArray
```
caroott, line 405: `tableIds` lost its `useMemo`

 this now builds a fresh array on every render and hands it to `DndKit.SortableContext` as `items`.

dnd-kit keys its internal memos off `items` identity, so every parent render invalidates the sortable context and re-renders every tab, including while a drag is in progress. The memo that was removed probably existed for this reason.

src/Components/src/Page/ArcFileEditor/ArcFileFooterTabs.fs
```diff
@@ -293,34 +293,31 @@ type ArcFileFooterTabs =
...
+            let result = update nextArcFile
+            arcFileRef.current <- nextArcFile
+            setArcFile nextArcFile
+            result
 
         let isEditorModeTableTab, setIsEditorModeTableTab =
             React.useState (None: int option)
 
-        let arcFileHash = arcFile.GetHashCode()
+        let fileType = arcFile.RelatedArcFilesDiscriminate |> unbox<string>
```
caroott, line 317: `metadataTabLabel` now depends on a Fable only representation

`arcFile.RelatedArcFilesDiscriminate |> unbox<string>` only works because `ArcFilesDiscriminate` is a `[<StringEnum>]`, which holds under Fable and not under .NET.

The `CompiledName` values are all lowercase, so the capitalisation produces exactly the same labels as the old match, including `Datamap`. But `Swate.Components.fsproj` is also referenced by the .NET Expecto project where that cast is meaningless, and the old match was compiler checked and about the same length. Why not keep it?

Issue #1275, opened 2026-08-03
[Bug]: Cannot delete file from file explorer


Brilator:
### What happened?

Cannot delete file from file tree as it is half-recognized as assay. 

"Could not delete ARC entity at 'assays/.DS_Store': Cannot parse assay identifier from FileName `assays/.DS_Store/isa.assay.xlsx`"

https://github.com/user-attachments/assets/8528ef66-2f8e-429a-9854-6302d70e397c

SwateApp

PR #1309, opened 2026-08-07
Bug/fix empty file name deletion

Etschbeijer:
* Fix path finding for files and folders under entity collectors

### DIFF
--- a/src/Shared/PathHelpers.fs
+++ b/src/Shared/PathHelpers.fs
@@ -212,7 +212,8 @@ module ArcEntityPathRules =
     let private classifyStructuralArcPath =
         function
         | [| zoneSegment |] when isAddZoneSegment zoneSegment -> StructuralArcPath.AddZoneRoot
-        | [| zoneSegment; _ |] when isAddZoneSegment zoneSegment -> StructuralArcPath.EntityFolder
+        | [| zoneSegment; name |] when isAddZoneSegment zoneSegment && not (name.StartsWith ".") ->
+            StructuralArcPath.EntityFolder
         | [| zoneSegment; _; childFolderName |] ->
             match tryParseZone zoneSegment with
             | Some zone when PathHelpers.pathMatchesAny (nativeEntityChildFolderNames zone) childFolderName ->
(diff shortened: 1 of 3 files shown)


Etschbeijer:
Closes issue #1275

Gitarackur (dismissed):

src/Shared/PathHelpers.fs
```diff
@@ -293,6 +294,8 @@ module ArcEntityPathRules =
                 | None -> DeletePathClassification.GenericTarget normalizedRelativePath
             | [| zoneSegment; identifier |] ->
                 match tryParseZone zoneSegment with
+                | Some zone when identifier.StartsWith "." ->
```
caroott: delete is fixed but rename still misclassifies the same paths

`classifyDeleteTarget` learned about hidden entries but `classifyRenameTarget` did not.

`renamePath` in `IArcVaultsApi.fs` only routes `GenericTarget` to the generic filesystem rename, so this goes into `renameArcEntityAsync` and fails at `tryEnsureArcEntityResolved` with `ARC does not contain assay with identifier '.DS_Store'.`

So after this PR you can delete `assays/.DS_Store` but still cannot rename it, and the two classifiers disagree about what the same path is. The same one line guard in the `[| zoneSegment; identifier |]` arm of `classifyRenameTarget` fixes it.

src/Shared/PathHelpers.fs
```diff
@@ -212,7 +212,8 @@ module ArcEntityPathRules =
     let private classifyStructuralArcPath =
         function
         | [| zoneSegment |] when isAddZoneSegment zoneSegment -> StructuralArcPath.AddZoneRoot
-        | [| zoneSegment; _ |] when isAddZoneSegment zoneSegment -> StructuralArcPath.EntityFolder
+        | [| zoneSegment; name |] when isAddZoneSegment zoneSegment && not (name.StartsWith ".") ->
```
caroott: Every other comparison in this module goes through `PathHelpers.pathsEqual` or an explicit `StringComparison`. On .NET, `string.StartsWith(string)` defaults to `CurrentCulture`.

Use the char overload `name.StartsWith '.'` or `name.StartsWith(".", System.StringComparison.Ordinal)` to match the rest of the file.

src/Shared/PathHelpers.fs
Etschbeijer: Fixed in commit: 1951c84292ea0e9c49171e8cb94dea21add274a1

src/Shared/PathHelpers.fs
Etschbeijer: Fixed in commit: 1951c84292ea0e9c49171e8cb94dea21add274a1

caroott (approved):

Issue #1286, opened 2026-08-04
[Feature Request]: DataHUB wording


Brilator:
### Is your feature request related to a problem?

related to #1283, but specifically for DataHUB

<img width="956" height="318" alt="Image" src="https://github.com/user-attachments/assets/1193cca4-5ef1-4c3b-82f7-da19f29cfe15" />

- unify datahub capitalization to DataHUB
- "Organisations" on GitLab / DataHUB are called "Groups"
- Instead of GitLab Explore -> "DataHUB explore" or actually "Download ARC from DataHUB" as the function button on the left suggests

### Host

Applies to All

Issue #1283, opened 2026-08-04
[Feature Request]: Adapt Git view / wording


Brilator:
### Is your feature request related to a problem?

Current implementation of Git / source control is a mix of using vs. abstracting git terminology and functions from users. IMO most important to the users is the control / awareness of what is shared "online" and when. Hence the ARCitect's split into views for "Commit" (local) and "DataHUB sync" (remote); although I personally also prefer the current more condensed all-in-one view.

<img width="569" height="460" alt="Image" src="https://github.com/user-attachments/assets/be8f57a5-e49f-4021-bd71-5a20dae38b00" /> 

### Describe the solution you would like!

Some suggestions: 
- unify wording "Git" (sidebar) and "Source control" (top headline)
- Rename "Update ARC from Online" -> Pull (and integrate) changes from DataHUB 
- Split the dropdown with "Save All Changes" and "Add and commit all changes" into two always available buttons: 
  1. "Commit and push all changes to the DataHUB"
  2. "Commit all changes locally" 
  (ignore the  "add" part...) 
- subheadlines "Actions" and "Save" could be removed without loosing much 
### Host

SwateApp

caroott:
The naming with Save instead of Commit was intentional, are we sure we want to drop it?
Also, Source Control as unified wording would probably be better than git, since we plan to potentially use other versioning (e.g. LakeFS) in the future.
We are also not only supporting DataHUB here as far as i know. Any gitlab can be a configured source. Maybe a third name would be better than either of the two before?

Brilator:
I don't know. "Versions / Version Control" then? 
In any case, it must be user-comprehensible ("Source Control" is not).

Brilator:
Also, I would like to see and edit the remote URL. 

- [ ] Show remote URL 
- [ ] Make remote URL editable

PR #1311, opened 2026-08-07
Update and unify descriptions

Etschbeijer:
* Updated and unified descriptions in the UI


Etschbeijer:
* Close issue #1283
* Close issue #1286

caroott:
The naming with Save instead of Commit was intentional, are we sure we want to drop it?
Also, Source Control as unified wording would probably be better than git, since we plan to potentially use other versioning (e.g. LakeFS) in the future.

caroott:
We are not only supporting DataHUB here as far as i know. Any gitlab can be a configured source. Maybe a third name would be better than either of the two before?

Issue #1305, opened 2026-08-06
[Feature Request]: Check for git config credentials


Brilator:
### Is your feature request related to a problem?

Users who just installed Git / Git-lfs for a workshop and have never used it, struggled to authenticate, commit, push, etc; since no username and email were git configured. 

related #1304 

### Host

Applies to All

PR #1315, opened 2026-08-10
Sign commits with the signed-in DataHub account

### LINKED ISSUE #1304: [Bug]: Git commits are not signed properly
### What happened?

When removing my system user git config credentials and only use the PAT authentication in Swate App, my commits on the ARC are signed with a wrong username (my display name, not my user name shorthand) which results in commits not properly linked to my account. 

### Host

Applies to All

### Host Details

_No response_

caroott:
Hopefully fixes #1304 & #1305

Commits made in Swate never carried the signed-in account's identity. Git fell back to whatever it could derive from the OS, so commits showed up with the wrong author and never linked to the DataHub account (#1304). On machines without any git identity configured, committing failed with git's raw "Please tell me who you are" error (#1305).

This PR wires the account identity into every commit path:

- The GitLab username is now stored with each account and used as the commit author name.
- Commits, merge resolution commits and pull merge commits all run with the account's `user.name` and `user.email` passed as scoped config, the same way the access token is already handled. Nothing is written to `.git/config`, and with no account signed in the behavior is unchanged, so a locally configured identity still works.
- Accounts with GitLab's "use a private email in commits" setting enabled commit with their noreply address instead of exposing their primary email.
- If git still can't resolve an identity, the error is now classified and shown with actual setup guidance instead of git's raw output.

Existing stored accounts keep working. The new fields are optional in the metadata store and fill themselves in on the next token revalidation.


src/Electron/src/Main/Git/GitService.fs
```diff
@@ -1058,8 +1105,12 @@ let private createAuthenticatedGitSession
 
                     match tokenOption with
                     | Some token when not (String.IsNullOrWhiteSpace token) ->
+                        let! commitIdentityEntries = resolveCommitIdentityConfigEntries ()
```
Etschbeijer: When multiple accounts are active, there is a chance that two different accounts are used or not?
tryGetAccessToken host could access a different account than resolveCommitIdentityConfigEntries() or not?

Etschbeijer (commented):
lgtm

Etschbeijer (approved):

PR #1319, opened 2026-08-13
Remove extra column space

Etschbeijer:
* Remove the extra column space, so that the focus no longer jumps when the last colum is selected


Freymaurer (changes requested):
You have not explained the issue in the PR, you have ignored my comment in the code and just removed the extra whitespace without getting feedback :D

PR #1325, opened 2026-08-18
Fix overscroll bug with scrolling to last column/row with whitespace …

Freymaurer:
# The Bug

1. Have the window smaller than the whitespace in the Table component, then have multiple columns to allow scrolling.
2. Use either keyboard navigation or mouseclicks to navigate to the last column.
3. ⚠️ the last column MUST NOT be fully visible or the bug is not triggered. This is most easily done by doing it with keyboard navigation
4. During the transition to the last column, instead of scrolling the last column into view, tanstack virtual will scroll to the scroll container element end 🐛 This is a bug in tanstack virtual, tracked here: https://github.com/TanStack/virtual/issues/1257

## What i did

- Implement a hotfix, forcing a different scroll behavior for the last index, by using `align=center`, which changes scroll behavior. This is not optimal, as it will not scroll the element into view, instead it will position it in the view center. **This is different behavior to earlier scrolls**, but acceptable.
- Create minimal reproduction and issue in tanstack virtual to have this solved at the source of the issue
- add a comment to my bandaid fix, with reference to the issue and description.

## What i tried

- calculating the exact position of the element and scrolling there. This is not possible as tanstacks `_.getVirtualItem` only returns the currently renderer items (so like 12 items). This will propably work for keyboard navigation, but will not work if we at some point implement a search function and are scrolling across the table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Report a missing git identity as its own failure with setup guidance instead of passing git's raw "Please tell me who you are" output to the user #1305.
 -   Respect GitLab's "use a private email in commits" setting by signing commits with the account's commit email instead of its primary email.
 -   Fix Template bug "You've hit dummy code" #1266, #1264 (by @Freymaurer) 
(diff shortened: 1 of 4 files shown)


Etschbeijer (dismissed):

Etschbeijer (approved):
Lgtm

PR #1328, opened 2026-08-19
start with avpr

### LINKED ISSUE #127: [Feature Request] Checksum as Checklist type
**Is your feature request related to a problem? Please describe.**
Some repositories require a specific type of checksum for your raw data files (e.g. GEO requires MD5). 

**Describe the solution you'd like**
I know this is really shooting for the stars here... 
Implement MD5, sha256, etc. as validation type.

Freymaurer:
closes #127

image

---

## Inputs

- `config: ValidationPackagesConfig`, the existing config, read from disc or if none exists initilized with empty.
- `writeConfig: ValidationPackagesConfig -> JS.Promise<Result<unit, exn>>`, will update in-memory state of config and write to disc
- `fetchValidationPackages: unit -> JS.Promise<ValidationPackageDTO[]>`, implemented with [avpr api call](https://avpr.nfdi4plants.org/swagger/index.html#/Validation%20Packages/GetAllPackages) in mind. Used to get all available validation packages

## Features

- Display all available validation packages with some metadata information. Reduce column count on smaller screens. Paginated results, 20 items per page.
- "Legacy" check. If a package name in config cannot be found in available packages, then it is listed in a warning banner. Opening the banner will sow a list of effected packages with a single click option to remove them
- "outdated" check. If a package has a lower version than the one available online, then there will be a "Update" button next to the version. 
- filter in header for tags and authors. allowing filter for specific metadata entries.
- sort button for checked state in checkbox header
- search input with "config", allowing to search specific metadata fields, such as name, summary, description.
- Changes are only applied via Submit button (This might be subject to change in the future, maybe unified to a generalized page layout)

---

Search input config

image

Update button for older versions

image

Legacy package list

image

Filter option

image


src/Components/src/Composite/ValidationPackageSelector/Context.fs
```diff
@@ -0,0 +1,29 @@
...
+
+type ValidationPackageSelectorContext = {
+    FetchState: SelectorState
+    Packages: ValidationPackageDTO[]
+    RowStateOf: ValidationPackageDTO -> PackageRowState
+    Toggle: ValidationPackageDTO -> unit
+    UpdateToLatest: ValidationPackageDTO -> unit
+}
+
+module ValidationPackageSelectorContextHelper =
```
Etschbeijer: Do you really need a helper module for this? I think a normal function call would be enough for the initial function or not?

src/Components/src/Composite/ValidationPackageSelector/PackagePagination.fs
```diff
@@ -0,0 +1,65 @@
...
+            Html.tfoot [
+                Html.tr [
+                    Html.td [
+                        prop.colSpan 8
+                        prop.className "swt:bg-base-100"
+                        prop.children [
+                            Html.div [
+                                prop.className "swt:flex swt:items-center swt:justify-center swt:gap-6 swt:py-2"
+                                prop.children [
+                                    Html.button [
```
Etschbeijer, line 54: Maybe you can use an existing button component? That would make it a bit easier to overlook the react element.

src/Components/src/Composite/ValidationPackageSelector/SearchField.fs
```diff
@@ -0,0 +1,88 @@
...
+open Swate.Components.Composite.ValidationPackageSelector.Context
+open Swate.Components.Primitive.Popover
+open Types
+
+[<Erase; Mangle(false)>]
+type SearchField =
+
+    [<ReactComponent>]
+    static member private ScopeCheckbox(field: SearchFields, isChecked: bool, onToggle: SearchFields -> unit) =
+        Html.label [
```
Etschbeijer, line 15: We use toggle quite often. I think at this point it might be a good idea to write a component for it.

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
```diff
@@ -0,0 +1,218 @@
...
+            React.useMemo ((fun () -> Helper.distinctTags items), [| box items |])
+
+        let authorOptions =
+            React.useMemo ((fun () -> Helper.distinctAuthors items), [| box items |])
+
+        let tagFilterName =
+            selectedTags
+            |> Set.toList
+            |> List.tryHead
+            |> Option.map (fun i -> tagOptions.[i])
```
Etschbeijer: "selectedTags" stores the selected indices from "tagOptions". 
If "tagOptions" changes when the search changes, those stored indices may no longer refer to the same tags.
This could make an index point to a different option or go out of range at |> Option.map (fun i -> tagOptions.[i]).

src/Components/src/Composite/ValidationPackageSelector/ValidationPackageSelector.fs
```diff
@@ -0,0 +1,274 @@
...
+        let isDirty =
+            React.useMemo (
+                (fun () ->
+
+                    let newConfig =
+                        ValidationPackageSelectorModel.createNextConfig config packages edits removedUnlisted
+
+                    newConfig <> config_old
+                ),
+                [| box edits; box removedUnlisted |]
```
Etschbeijer: Why are newConfic, config_old and packages not included here? At first glance it looks like it could become stale.

Etschbeijer (changes requested):

src/Components/src/Composite/ValidationPackageSelector/PackagePagination.fs
Freymaurer: We could create a composite PaginatedTable component soon, but replacing this with primitive buttons seems overengineered

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
Freymaurer: Good catch! This case actually triggers by selecting a tag/author and then using the searchbar. as the available tags are calculated after the searchbar. 

I implemented a `useKeyedState` hook:

```fsharp
[<Hook>]

    static member useKeyedState<'value, 'key when 'key: equality>(initialValue: 'value, key: 'key) =
        let state, setState = React.useState (KeyedState.init(initialValue, key))

        let value =
            if state.key = key then
                state.state
            else
                initialValue

        let setValue (value: 'value) =
            setState (KeyedState.init(value, key))

        let setValueStable = React.useCallback (setValue, [| key |])

        value, setValueStable
```

input are a initialValue and a key. Now whenever you set the state you update the internal key with the current one. No whenever another state changes the calculated key, we instead return the initial value. 

For this example with tags, this will look like this:

```fsharp
let tagOptions = Helper.distinctTags items
let tagsKey = tagOptions |> Array.map (fun t -> t.ToLower()) |> String.concat ","
let selectedTag, setSelectedTag = React.useKeyedState<int option, string> (None, tagsKey)
```

Should the available tags change, the key changes and the selected state returns the initial value `None`. Then if the state is set again, it gets the new key and works as intended.

This is a very robust and minimal implementation without any side effects. The only "feature/bug" is, that the state stores the old key with value, so if the key returns to the old value we go back to the prev state again. 

This means: Let us assume 2 datasets. We start with dataset one and add a keyedState for row selection, with the key being dataset.id. We then select a row. If we change dataset, the keyedState returns initial value None. If we change back to the previous dataset, the old key is restored and we revert back to the previously selected row.

src/Components/src/Composite/ValidationPackageSelector/SearchField.fs
Freymaurer: i mean, yes it could make sense to implement Primitives for all, well, primitive components. But this does only make sense for specific repeating patterns, such as "CloseButton" or if we actually go for full accessibility best practises. For now i would not do this.

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
Freymaurer: nevermind. i changed the component structure and now its not relevant anymore

src/Components/src/Composite/ValidationPackageSelector/SearchField.fs
Etschbeijer: answered

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
Etschbeijer: Fixed

src/Components/src/Composite/ValidationPackageSelector/Helper.fs
```diff
@@ -0,0 +1,201 @@
...
+    |> fun suffix -> baseVersion + suffix
+
+let rowState (config: ValidationPackagesConfig) (dto: ValidationPackageDTO) =
+    config.ValidationPackages
+    |> Seq.tryFind (fun (vp: ValidationPackage) -> vp.Name = dto.Name)
+    |> Option.map (fun vp ->
+        if vp.Version = Some(toVersionString dto) then
+            PackageRowState.Checked
+        else
+            PackageRowState.HasOlderVersion
```
Etschbeijer: When I understand this correctly, very non-equal version gets classified as “older,” including None and potentially a version newer than the registry version, right?

How are versionless packages handled? They appear as HasOlderVersion, or not?

src/Components/src/Composite/ValidationPackageSelector/ValidationPackageSelector.fs
```diff
@@ -0,0 +1,321 @@
...
+                    |> Helper.filterByAuthor selectedAuthorFilter
+                    |> Helper.sortByChecked checkedSort rowStateOf
+                ),
+                [|
+                    box packages
+                    box searchFields
+                    box searchQuery
+                    box selectedTagFilter
+                    box selectedAuthorFilter
+                    box checkedSort
```
Etschbeijer, line 215: "rowStateOf" depends on "edits/config", but neither of them is included in this memo’s dependency list.

When the sorting is active before the checkbox is changed, then "filteredPackages" will not be recomputed, right?

src/Components/src/Primitive/Select/MultiSelect.fs
```diff
@@ -0,0 +1,93 @@
...
+            setSelectedIndices: Set<int> -> unit,
+            ?onSelect: int option -> unit,
+            ?triggerRenderFn: {| isOpen: bool |} -> ReactElement,
+            ?optionRenderFn: SelectItemRender<'a> -> ReactElement,
+            ?dropdownPlacement: FloatingUI.Placement,
+            ?middleware: FloatingUI.IMiddleware[],
+            ?showSelectAll: bool
+        ) =
+        let leadingItem =
+            match showSelectAll with
```
Etschbeijer, line 78: Some Select components are now missing the Select All option.
Template filter
Notesearch filter
Termsearch ConfigSetter

Main:

Image

feature/avpr-integration:

Image

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
```diff
@@ -0,0 +1,174 @@
...
+            ]
+        | SelectorState.Error e ->
+            Html.tbody [
+                Html.tr [
+                    Html.td [
+                        prop.colSpan 8
+                        prop.style [ style.textAlign.center ]
+                        prop.text "Error loading packages :("
+                    ]
+                    Html.details [
```
Etschbeijer: I think you cannot / shall not use details in <tr>.
I thin you should move it into Html.td because it can cause rendering problems otherwise, or not?

Etschbeijer (changes requested):

src/Components/src/Composite/ValidationPackageSelector/Helper.fs
Freymaurer: Doing actual SemVer comparison is quite the pain. I opened an issue in ARCtrl for a `IComparable` implementation and added a minimal implementation:

```fsharp
type SemVer with

    /// Compares two SemVer instances based on their pre-release identifiers.
    /// Returns:
    /// - `-1` if v1 < v2
    /// - `1` if v1 > v2
    /// - `0` if they are equal.
    static member comparePreRelease (v1: ARCtrl.Helper.SemVer.SemVer) (v2: ARCtrl.Helper.SemVer.SemVer) =
        match v1.PreRelease, v2.PreRelease with
        | None, None -> 0
        | Some _, None -> -1
        | None, Some _ -> 1
        | Some pre1, Some pre2 ->
            let pre1Parts = pre1.Split('.')
            let pre2Parts = pre2.Split('.')

            let rec compareParts parts1 parts2 =
                match parts1, parts2 with
                | [], [] -> 0
                | [], _ -> -1
                | _, [] -> 1
                | p1 :: rest1, p2 :: rest2 ->
                    match System.Int32.TryParse(p1), System.Int32.TryParse(p2) with
                    | (true, int1), (true, int2) -> 
                        if int1 < int2 then -1
                        elif int1 > int2 then 1
                        else compareParts rest1 rest2
                    | (false, _), (false, _) -> 
                        if p1 < p2 then -1
                        elif p1 > p2 then 1
                        else compareParts rest1 rest2
                    | (true, _), (false, _) -> -1
                    | (false, _), (true, _) -> 1

            compareParts (List.ofArray pre1Parts) (List.ofArray pre2Parts)

    /// Compares two SemVer instances to determine if v1 is older than v2.
    /// Required until actual IComparable implementation is done for SemVer (https://github.com/nfdi4plants/ARCtrl/issues/639)
    static member isOlder (v1: ARCtrl.Helper.SemVer.SemVer) (v2: ARCtrl.Helper.SemVer.SemVer) =
        match v1, v2 with
        | v1, v2 when v1.Major < v2.Major -> true
        | v1, v2 when v1.Major > v2.Major -> false
        | v1, v2 when v1.Minor < v2.Minor -> true
        | v1, v2 when v1.Minor > v2.Minor -> false
        | v1, v2 when v1.Patch < v2.Patch -> true
        | v1, v2 when v1.Patch > v2.Patch -> false
        | v1, v2 when v1.Major = v2.Major && v1.Minor = v2.Minor && v1.Patch = v2.Patch ->
            let preReleaseComparison = ARCtrl.Helper.SemVer.SemVer.comparePreRelease v1 v2
            preReleaseComparison < 0
        | _ -> false

    static member isEqualWithoutBuild(v1: ARCtrl.Helper.SemVer.SemVer) (v2: ARCtrl.Helper.SemVer.SemVer) =
        v1.Major = v2.Major &&
        v1.Minor = v2.Minor &&
        v1.Patch = v2.Patch &&
        v1.PreRelease = v2.PreRelease
```

Which is now used in `rowStateOf`, for which i added more inline comments to explain reasoning:

```fsharp
let rowState (config: ValidationPackagesConfig) (dto: ValidationPackageDTO) =
    let configPackage = config.ValidationPackages |> Seq.tryFind (fun vp -> vp.Name = dto.Name)
    match configPackage with
    | Some p ->
        match p.Version with
        | Some v ->
            let configSemVer = ARCtrl.Helper.SemVer.SemVer.tryOfString v
            let dtoSemVer = ARCtrl.Helper.SemVer.SemVer.tryOfString (toVersionString dto)
            match configSemVer, dtoSemVer with
            | _ , None ->
                // If the version in the DTO is not a valid SemVer, we consider it unchecked, and log an error
                Browser.Dom.console.error ($"Invalid SemVer in DTO for package {dto.Name}: {toVersionString dto}")
                PackageRowState.Unchecked
            | None, _ ->
                // If the version in the config is not a valid SemVer, we consider it unchecked
                PackageRowState.Unchecked
            | Some configSemVer, Some dtoSemVer ->
                if SemVer.isEqualWithoutBuild configSemVer dtoSemVer then
                    PackageRowState.Checked // The versions are equal, so the package is checked
                elif SemVer.isOlder configSemVer dtoSemVer then
                    PackageRowState.HasOlderVersion // The config version is older than the DTO version, so the package has an older version
                else
                    PackageRowState.InvalidVersion // The config version is newer than the DTO version, which is unexpected
        // To my understanding no version means, always take latest, so we can consider it checked
        // I think it might be best if Swate does list specific versions, so we will never actually write this case.
        | None -> PackageRowState.Checked 
    | None ->
        // Not found in current config 
        PackageRowState.Unchecked 
```

src/Components/src/Composite/ValidationPackageSelector/ValidationPackageSelector.fs
Freymaurer: To be honest, not sure. I abstracted a more simple map alyer ont op of it:

```fsharp
let RowStateMap =
    React.useMemo (
        (fun () ->
            packages
            |> Array.map (fun dto -> dto.Name, rowStateOf dto)
            |> Map.ofArray
        ),
        [| box packages; box edits; box config |]
    )

let filteredPackages =
    React.useMemo (
        (fun () ->
            packages
            |> Helper.filterBySearch searchFields searchQuery
            |> Helper.filterByTag selectedTagFilter
            |> Helper.filterByAuthor selectedAuthorFilter
            |> Helper.sortByChecked checkedSort RowStateMap
        ),
        [|
            box packages
            box searchFields
            box searchQuery
            box selectedTagFilter
            box selectedAuthorFilter
            box checkedSort
            box RowStateMap
        |]
    )
```

This now has clear dependencies and propagates correctly. This change also allowed me to more easily deploy the smart-controller/dumb-visual component pattern for PackagePagination (smart controller) and MemoizedPackageRow (dumb visual component)

```fsharp
[<ReactMemoComponent(AreEqualFn.FsEqualsButFunctions)>]
    static member MemoizedPackageRow(pkg: ValidationPackageDTO, rowState: PackageRowState, updateToLatest, toggle, ?key: obj) =
        PackageRow.PackageRow(pkg, rowState, updateToLatest, toggle, ?key = key)

for pkg in pageItems do
                    let rowState = ctx.RowStateMap |> Map.tryFind pkg.Name |> Option.defaultValue PackageRowState.Unchecked
                    PackagePagination.MemoizedPackageRow(pkg, rowState, ctx.UpdateToLatest, ctx.Toggle, key = pkg.Name)
```

Now a row only know what it must know, and does not rerender all the time when context changes.

src/Components/src/Primitive/Select/MultiSelect.fs
Freymaurer: upps! good catch! auto complete gave me the wrong defaults:

Now default is "on"

```fsharp
let leadingItem =
  match showSelectAll with
  | None 
  | Some true -> Some(MultiSelect.SelectAll(setSelectedIndices, "multi-select-select-all"))
  | Some false -> None
```

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
Freymaurer: good catch! silly mistake! I moved the details part into the td element:

```fsharp
| SelectorState.Error e ->
    Html.tbody [
        Html.tr [
            Html.td [
                prop.colSpan 8
                prop.children [
                    Html.div [
                        prop.className "swt:text-error swt:text-sm swt:font-semibold swt:flex swt:flex-col swt:gap-2"
                        prop.children [
                            Html.h1 "Error loading packages :("
                            Html.details [
                                prop.children [ Html.summary "Details"; Html.pre [ prop.text e.Message ] ]
                            ]
                        ]
                    ]
                ]
            ]
        ]
    ]
```

image

src/Components/src/Composite/ValidationPackageSelector/Helper.fs
Etschbeijer: lgtm

src/Components/src/Primitive/Select/MultiSelect.fs
Etschbeijer: lgtm

src/Components/src/Composite/ValidationPackageSelector/PackageTable.fs
Etschbeijer: lgtm

src/Components/src/Composite/ValidationPackageSelector/Helper.test.fs
```diff
@@ -0,0 +1,438 @@
...
+        Vitest.test (
+            "is Checked when version matches",
+            fun () ->
+                Vitest
+                    .expect(rowState (mkConfig [| "A", Some "1.0.0" |]) (ValidationPackageDTO.mkDefault (name = "A")))
+                    .toEqual
+                    PackageRowState.Checked
+        )
+
+        Vitest.test (
```
Etschbeijer, line 165: Can we add a test for a configured version newer than the AVPR version, e.g. 2.0.0 vs 1.0.0, and assert that this results in InvalidVersion?

Etschbeijer (dismissed):

src/Components/src/Composite/ValidationPackageSelector/Helper.test.fs
Freymaurer: yeah no worries, i added a test

PR #1329, opened 2026-08-19
Cancel git operations

### LINKED ISSUE #1306: [Feature Request]: kill git process
### Is your feature request related to a problem?

If for some reason, I get stuck with git processes (e.g. downloading an ARC) I would like to be able to kill the process. 

Image

### Describe the solution you would like!

_No response_

### Host

Applies to All

### Host Details

_No response_

caroott:
Closes #1306

Adds a cancel button for running git operations (ARC download, fetch, pull, push, LFS transfers). It kills the git process directly, so you're no longer stuck waiting when a download hangs.

Cleanup after canceling is handled so the ARC never ends up dirty: a canceled clone deletes the partial folder, a canceled pull aborts a half-applied merge/rebase and removes stale locks. You get a small "cancelled" notice instead of an error popup.

Internally this reuses the push cancellation we already had and extends it to all remote operations via simple-git's abort signal.

1-cancel-during-fetch
2-operation-cancelled-warning
3-cancel-during-lfs-download
4-pull-completed-lfs-cancelled-warning


Etschbeijer:
Add images, please

src/Electron/src/Renderer/Context/GitWorkflow.fs
```diff
@@ -270,6 +270,15 @@ let staleMergeConflictTokens = [|
     "changed on disk since it was opened"
 |]
 
+// Mirrors the token set GitService uses to classify GitFailureKind.Canceled, so cancelled
+// operations surface as a warning notice instead of an error dialog.
+let cancellationMessageTokens = [| "cancel"; "abort" |]
```
Etschbeijer: This could capture error messages that contain cancel or abort or not?
So, we never receive those messages

src/Electron/src/Main/Git/GitProvisioningService.fs
```diff
@@ -418,7 +452,10 @@ let cloneRepository
                                                             createCloneLfsCommandAuth host tokenOption safeRemoteUrl
 
                                                         let! hydrateResult =
-                                                            hydrateClonedLfsContent normalizedTargetPath commandAuth
+                                                            hydrateClonedLfsContent
+                                                                normalizedTargetPath
+                                                                commandAuth
+                                                                (Some cancellationScope.CancelCheck)
```
Etschbeijer, line 458: Git pull has finished when this starts, or not?
What happens when I cancel it before the hydration is finished? That would create an error / cancel request but we already moved at least partly into the new commit(s) or not?

Etschbeijer (changes requested):

src/Electron/src/Renderer/Context/GitWorkflow.fs
caroott: true, "Aborting" from git itself would land in there. dropped the token matching, we check the failure kind now instead

src/Electron/src/Main/Git/GitProvisioningService.fs
caroott: clone is fine since we delete the folder on cancel. for pull yes, the commits are already there and I don't think we should roll them back. now it just says pull worked but lfs download was cancelled, so the pointers stay until next update

src/Electron/src/Renderer/Context/GitWorkflow.fs
Etschbeijer: looks good

src/Electron/src/Main/Git/GitProvisioningService.fs
Etschbeijer: Solved

Etschbeijer (approved):
lgtm

Issue #1288, opened 2026-08-04
[Feature Request]: import folders / files via context menu to ARC


Brilator:
e.g. protocols, dataset files, etc.

### Host

SwateApp

Etschbeijer:
Will be released with 2.2.0

PR #1332, opened 2026-09-01
Feature/add import files

Etschbeijer:
# Description

* This PR adds external file importing to the FileTree in Electron because that functionality was lacking till now.

* Users can now import one or multiple files directly into an ARC folder through the FileTree context menu. 
* While files are being imported, the application displays the import state and allows the operation to be cancelled.
* The underlying Electron handling has been extended so imports are performed safely without overwriting existing files or leaving partial files behind after cancellation or failure. 
* Imported ARC files are also synchronized with the in-memory ARC, while file watcher and explicit import merges are processed sequentially to avoid conflicting updates.
* Additionally, native taskbar progress is shown during imports and regression tests were added for the import workflow, cancellation and cleanup, context-menu behaviour, and sequential ARC merge processing. 
* No new external dependencies are required.

* https://github.com/nfdi4plants/Swate/issues/1288

## Type of change

Added feature

- [x] New feature (non-breaking change which adds functionality)

- [x] Components
- [x] Electron - Renderer
- [x] Electron - Main

# How Has This Been Tested?

* [x] File import: Imports an external file into the selected ARC folder
* [x] Import cancellation: Cancelling removes imported and temporary files
* [x] Import failure: Failed imports clean up temporary and partially imported files
* [x] Existing files: Imports do not overwrite files already present in the target folder
* [x] Context menu: Import files is available for folders and targets the selected folder
* [x] Merge sequencing: Concurrent ARC merges are processed sequentially
* [x] Merge recovery: A failed ARC merge does not block following merge operations

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

> The Zen of GitHub.

- Practicality beats purity
- Speak like a human
- Encourage flow
- Favor focus over features
- Design for failure
- Keep it logically awesome


src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -95,6 +122,21 @@ let private runLoadedArcPathAction
             return Error e
     }
 
+let private pickAbsolutePaths (event: IpcMainInvokeEvent) = promise {
```
caroott: is this used somewhere?

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -203,6 +206,183 @@ let mapRenameDiskError (sourcePath: string) (targetPath: string) (renameError: e
...
+            let destinationPath = join [| targetDirectory; entry.FileName |]
+            let! destinationExists = pathExistsAsync destinationPath
+
+            if destinationExists then
+                raise (exn $"Cannot import '{entry.FileName}' because a file with that name already exists.")
+
+        return {
+            TargetDirectory = targetDirectory
+            // The watcher explicitly ignores this operation-owned directory and its contents.
+            TemporaryDirectory = join [| targetDirectory; $".swate-import-{Guid.NewGuid():N}" |]
```
caroott: Fable does not handle the `:N` format specifier inside interpolated strings. The compiled output for this line is the literal `` `.swate-import-%P(N)` ``. No GUID at all.

That breaks two things. The watcher ignore rule in ArcVaultHelper.fs expects 32 hex characters after the prefix, so it never matches and the staging folder plus every temp copy is watched during the import. And because the name is constant, a killed import leaves `.swate-import-%P(N)` behind and the next import into the same folder silently reuses it (`mkdirAsync` is recursive).

The tests use the same for their file names (`cancel-first-%P(N).txt` in the compiled test) and only assert on the `.swate-import-` prefix, so they pass without noticing.

`Guid.NewGuid().ToString("N")` works. I'd also add a test that the generated name satisfies the ignore predicate

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -118,17 +200,16 @@ module ArcVaultExtensions =
...
                         // ARC.merge may copy or reconstruct entities without retaining the hashes established
                         // above, so transfer the disk baseline to the merged ARC. Watcher-applied values then
                         // stay clean, while values that still differ because of local edits remain unsaved.
                         syncArcStaticHashes reloadedArc mergedArc
                         this.SetArc mergedArc
                         this.RefreshHasUnsavedArcChangesFlag()
-            | Some _, None -> do! this.LoadArc()
-            | None, _ -> ()
+                        return Ok()
+            | _ -> return Error(arcNotOpenError ())
```
caroott, line 120: This drops the `Some path, None -> LoadArc()` branch. `Startup` starts the watcher before `LoadArc`, so a vault whose initial load fails is left with `path = Some`, `arc = None` and a live watcher. On main the next disk change loaded the ARC. Now every watcher batch logs "Unable to merge ARC after file watcher event" and nothing reloads it until the window is reopened.

`RenameOpenArcRoot` still special-cases `arc.IsNone`, so this state is clearly expected to happen. I'd keep the recovery branch.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -41,6 +57,81 @@ type ArcVault(window: BrowserWindow) =
...
+        let queuedMerge = promise {
+            try
+                do! precedingMerge
+            with precedingError ->
+                swatelogfn this.window.id "A preceding ARC merge failed: %s" precedingError.Message
+
+            return! operation ()
+        }
+
+        lastArcMerge <- promise {
```
caroott: `lastArcMerge` is a promise derived from `queuedMerge` with no rejection handler. If a merge rejects (as opposed to returning `Error`), the caller handles its own copy but this derived promise rejects unhandled in the main process until some later enqueue attaches its `try/with`. If that later enqueue never comes, it stays unhandled.

The "continues processing after a merge operation fails" test enqueues the second merge before the first rejects, so it doesn't hit this.

`lastArcMerge <- queuedMerge |> Promise.map ignore |> Promise.catch ignore` fixes it and the next enqueue already logs the error.

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -203,6 +206,183 @@ let mapRenameDiskError (sourcePath: string) (targetPath: string) (renameError: e
...
+            | Ok targetDirectory ->
+                let createdTargetPaths = ResizeArray<string>()
+                let mutable temporaryDirectory = None
+
+                try
+                    let! plan = createExternalFileImportPlan targetDirectory sourcePaths
+                    temporaryDirectory <- Some plan.TemporaryDirectory
+                    do! copyExternalFilesToTemporaryDirectory plan onProgress isCancellationRequested
+                    do! copyTemporaryFilesIntoTarget plan createdTargetPaths isCancellationRequested
+                    do! rmAsync plan.TemporaryDirectory (RmOptions(recursive = true, force = true))
```
caroott: At this point every file is already linked or copied into place, but removing the staging directory is still inside the `try` that owns rollback. `rm` with `force` only swallows ENOENT and there is no `maxRetries`, so a transient EBUSY or EPERM on Windows (indexer, antivirus holding the file) throws, the handler deletes the just-imported files and the user gets "Import failed" for an import that actually worked.

Cleanup after placement should be best effort. Log it, don't roll back.

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -203,6 +206,183 @@ let mapRenameDiskError (sourcePath: string) (targetPath: string) (renameError: e
...
+
+                    match! validateImportedFiles () with
```
caroott, line 398: The last `isCancellationRequested` check is inside `copyTemporaryFilesIntoTarget`. A cancel that arrives during `validateImportedFiles` (a full `ARC.LoadAsyncSwate` plus merge, which can take a while on a big ARC) is never read and the import returns `Completed`.

Meanwhile the renderer has already switched to "Cancelling import..." and hidden the button (FileTree.fs:590). So the UI says cancelled, the files stay. Same thing for a close-initiated cancel.

Either check the flag once more after validation and roll back, or don't claim cancellation in the UI once the copy phase is done.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -379,6 +420,79 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
...
+
+                                            let! result =
+                                                ArcFileSystemHelper.importExternalFilesOnDisk
+                                                    vault.path.Value
+                                                    request.targetRelativePath
+                                                    request.sourceAbsolutePaths
+                                                    reportImportProgress
+                                                    (fun () -> abortSignal.aborted)
+                                                    (fun () ->
+                                                        vault.TryTriggerArcInMemoryMergeOnFileWatcherEvents(
```
caroott: The import never sets `isBusyWriting`, unlike `AddArcFile`, `CreateARC`, delete and rename. Two things follow from that.

The watcher ignore only covers the staging folder, so chokidar sees the final files. They pass `IsFileWatcherArcMergeEligible`, and 500 ms later the whole ARC is loaded and merged a second time, and the tree is updated again on top of `refreshVaultFileTree`. With the GUID bug above the temp copies trigger a merge as well.

And nothing keeps a `WriteArc` from running at the same time as this validation merge. The watcher path checks `not this.isBusyWriting` before merging; this path doesn't.

Simplest option: run the import under `isBusyWriting` like the other app-initiated writes and drop one of the two tree refreshes.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -604,40 +691,58 @@ type ArcVaults() =
...
 
     member this.OnCloseWindow(window: BrowserWindow, vault: ArcVault, id: int) =
 
         window.onClose (fun closeEvent ->
-            if not vault.isCloseApproved then
-                if vault.hasUnsavedArcChanges then
+            match vault.CloseState with
+            | CloseLifecycleState.Approved -> ()
+            | CloseLifecycleState.WaitingForImportCleanup
+            | CloseLifecycleState.WaitingForSaveDecision -> closeEvent.preventDefault ()
```
caroott: `WaitingForSaveDecision` now calls `preventDefault` unconditionally and never re-requests the decision. On main a pending request still fell through to the unsaved-changes check and closed the window once the changes were gone.

Only `ResolveCloseRequest` leaves this state. If the renderer modal is lost (reload, crash), the window can't be closed any more. Unlikely, but who knows...

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
```diff
@@ -60,6 +61,13 @@ type FileTree =
 
         let activeDialog, setActiveDialog = React.useState<FileTreeDialog option> None
         let isDialogBusy, setIsDialogBusy = React.useState false
+
+        let activeImportRequestId, setActiveImportRequestId =
```
caroott: All the import state lives in this component: the request id, the overlay, the cancel button. LeftSidebar/Main.fs:23 unmounts FileExplorer when the sidebar switches to the Git page, so switching tabs mid-import drops all of it while main keeps copying.

Coming back mounts a fresh FileTree with no request id. The user can start a second import, main rejects it with "Another file import is already running", and there is no way left to cancel the first one from the UI.

The state needs to live above the sidebar switch, or main needs a way to cancel its active import without being handed the request id.

tests/Electron.Core/ArcFileSystemHelper.test.fs
```diff
@@ -10,6 +10,7 @@ open Swate.Electron.Shared.FileIOTypes
 open Vitest
 
 let private fsPromisesDynamic: obj = importAll "fs/promises"
+let private fsDynamic: obj = importAll "fs"
```
caroott: where is this used?

src/Electron/src/Main/Bindings/Filesystem.fs
```diff
@@ -71,6 +74,18 @@ let renameAsync (oldPath: string) (newPath: string) : JS.Promise<unit> = jsNativ
 [<Import("rm", "fs/promises")>]
 let rmAsync (path: string) (options: RmOptions) : JS.Promise<unit> = jsNative
 
+[<Import("copyFile", "fs/promises")>]
```
caroott, line 74: `copyFileAsync` and `copyFileWithFlagsAsync` both import `fs/promises.copyFile`. Node's `mode` argument defaults to 0, so one binding with a `mode` parameter covers both call sites.

src/Electron/src/Main/IPC/FileSystemIO.fs
```diff
@@ -203,6 +206,183 @@ let mapRenameDiskError (sourcePath: string) (targetPath: string) (renameError: e
...
+
+    type private ExternalFileImportEntry = { SourcePath: string; FileName: string }
+
+    type private ExternalFileImportPlan = {
+        TargetDirectory: string
+        TemporaryDirectory: string
+        Entries: ExternalFileImportEntry[]
+    }
+
+    let private resolveImportTargetDirectory arcPath targetRelativePath = promise {
```
caroott: `resolveImportTargetDirectory` repeats the empty-path-means-root branch from `resolveCreatePathPair` and the `directoryExistsAsync` / "is not a folder" guard from `createFileSystemItemOnDisk`. And `createExternalFileImportPlan` loops `pathExistsAsync` and raises where `ensureTargetDoesNotExist` already exists and returns a `Result`.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -15,16 +15,43 @@ open Swate.Electron.Shared.FileIOHelper
...
 open Main
+open Main.Bindings
 open Main.ARCtrlExtensions
 open Main.ArcVaultHelper
 open Main.IPC.Delete
 open Main.IPC.Rename
 open Swate.Electron.Shared.DTOs.ProvenanceGroupingDto
 open Main.IPC.FileSystemIO
 
+let private createImportedFileWatcherEvents arcPath (request: ImportExternalFilesRequest) =
```
caroott: `WatcherHelpers.buildWatcherEvent` already builds an `ArcVaultFileSystemEvent` from an absolute path and is what the real watcher uses. This function rebuilds the relative path by string concatenation. If path handling ever changes in the helper, this copy drifts.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -41,6 +57,81 @@ type ArcVault(window: BrowserWindow) =
...
+        match this.CloseState, this.activeFileImport with
+        | CloseLifecycleState.Idle, Some _ ->
+            Fable.Core.JS.Constructors.Promise.resolve (
+                Error(exn "Another file import is already running in this window.")
+            )
+        | CloseLifecycleState.Idle, None ->
+            let abortController = AbortController.create ()
+
+            let completion = promise {
+                do! Promise.sleep 0
```
caroott: The `Promise.sleep 0` isn't needed. Fable.Promise's `TryFinally` is implemented with `either`, which is a `.then` continuation, and `Delay` turns a synchronous throw into a rejection. The `finally` can't run before `activeFileImport` is assigned. Same goes for the `RequestId` comparison in the `finally`; `RunFileImport` refuses to start while an import is active, so the active one is always this one.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -379,6 +420,79 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
...
+                                                    reportImportProgress
+                                                    (fun () -> abortSignal.aborted)
+                                                    (fun () ->
+                                                        vault.TryTriggerArcInMemoryMergeOnFileWatcherEvents(
+                                                            importedEvents |> Array.toList
+                                                        )
+                                                    )
+
+                                            match result with
+                                            | Ok _
```
caroott: Both arms do the same thing, so this is just `do! refreshVaultFileTree vault`. Also, the refresh runs on `Cancelled` and `Error`, where cleanup has already put the tree back the way it was.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -15,16 +15,43 @@ open Swate.Electron.Shared.FileIOHelper
...
+            |> PathHelpers.normalizePath
+
+        {
+            EventName = Chokidar.Events.Add.ToString()
+            RelativePath = relativePath
+            AbsolutePath = Main.Bindings.Path.join [| arcPath; relativePath |]
+        }
+    )
+
+let private refreshVaultFileTree (vault: ArcVault) = promise {
```
caroott: Good helper, but lines 579, 811 and 884 in this same file still inline the two lines it replaces.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
```diff
@@ -543,6 +587,55 @@ type FileTree =
...
+                            prop.children [
+                                Swate.Components.Primitive.LoadingSpinner.LoadingSpinner.LoadingSpinner(
+                                    text =
+                                        if isCancellingImport then
+                                            "Cancelling import..."
+                                        else
+                                            "Importing files..."
+                                )
+                                if not isCancellingImport then
+                                    Swate.Components.Composite.AnnotationTable.FooterButtons.Cancel(
```
caroott: `Composite.AnnotationTable.FooterButtons.Cancel` has only ever been used by the AnnotationTable modals. Pulling it into the file explorer ties this overlay to a restyle of those footers. The Git sidebar's cancel button for in-flight operations is the closer precedent.

caroott (dismissed):

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commits: 4b5d3be and 2127e1e

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
Etschbeijer: Fixed in commits: 4b5d3be and 2127e1e

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: Fixed in commits: 4b5d3be and 2127e1e

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commits: 4b5d3be and 2127e1e

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: f6db88d263c26023ce91b36daf7223121f7918a8

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: Fixed in commit: f6db88d263c26023ce91b36daf7223121f7918a8

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: f6db88d263c26023ce91b36daf7223121f7918a8

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: Fixed in commit: f6db88d263c26023ce91b36daf7223121f7918a8

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: f6db88d263c26023ce91b36daf7223121f7918a8

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: No longer used. Removed it.

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
Etschbeijer: Fixed in commit: 3ae57f207d5152c102fa57a8ef85bde1c9001afb

tests/Electron.Core/ArcFileSystemHelper.test.fs
Etschbeijer: Removed, no longer used.

src/Electron/src/Main/Bindings/Filesystem.fs
Etschbeijer: Fixed in commit: 3ae57f207d5152c102fa57a8ef85bde1c9001afb

src/Electron/src/Main/IPC/FileSystemIO.fs
Etschbeijer: Fixed in commit: cd51f546d65d7b9125c0b5b1d6138619579d5eb0

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: cd51f546d65d7b9125c0b5b1d6138619579d5eb0

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: cd51f546d65d7b9125c0b5b1d6138619579d5eb0

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: cd51f546d65d7b9125c0b5b1d6138619579d5eb0

src/Electron/src/Renderer/Components/LeftSidebar/FileExplorer/FileTree.fs
Etschbeijer: 5e1787bc627e33f31ac0a18b3b9fec877fb376a9

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -245,8 +276,10 @@ module ArcVaultExtensions =
 
         /// Writes the active in-memory ARC scaffold to disk without touching unmanaged files such as notes.
         member this.WriteArc() : Fable.Core.JS.Promise<Result<unit, exn>> = promise {
-            match this.path, this.arc with
-            | Some arcPath, Some arc ->
+            match this.isBusyWriting, this.path, this.arc with
+            | true, _, _ ->
+                return Error(exn "Swate is still saving another change. Please wait a moment and try again.")
```
caroott, line 281: Why error and not busy until write successful? Users don't know how long the ARC is writing

src/Electron/src/Main/IPC/IPCHelper.fs
```diff
@@ -43,3 +43,14 @@ module IPCHelper =
...
                 vault.isBusyWriting <- false
         }
+
+    let withExclusiveBusyWriting
+        (vault: ArcVault)
+        (operation: unit -> JS.Promise<Result<'T, exn>>)
+        : JS.Promise<Result<'T, exn>> =
+        if vault.isBusyWriting then
+            JS.Constructors.Promise.resolve (
+                Error(exn "Swate is still saving another change. Please wait a moment and try again.")
```
caroott, line 53: Same

caroott (commented):
I can add files to locations where they are unexpected (e.g. assay folder locations). After i did that, i can't remove them anymore.

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: I wanted to avoid increasing the size of the PR further (I already removed parts because it became too big.)
'isBusyWriting' is at the moment only a flag because waiting would require a serialization so the queued operation reads the current ARC only after the previous write has completed.

Freymaurer (dismissed):

kMutagene (approved):

PR #1337, opened 2026-09-08
Fix a copy paste bug in annotation table and datamap

Etschbeijer:
* Fix a bug that disabled the recognition of column headers in the annotation table
* Fix a bug that moved the column by +1 when pasted in datamap
* Fix bug in datamap that prevented pasting of terms and units into datamap
* Refactor copy paste logic to streamline it and increase its robustness


Etschbeijer:
# Description

* **Clipboard:** Copy/paste handling was refactored around a shared structured clipboard format while retaining plain-text TSV interoperability. Structured cell data such as ontology terms and units is now preserved when copying and pasting within Swate, with browser-compatible fallbacks for clipboard formats that are not available.
* **Annotation Table:** Column-header recognition during paste was fixed. Paste target handling and related copy/paste utilities were also simplified to make selection behavior more consistent and robust.
* **DataMap:** Pasting no longer shifts the target column by one. Copy/paste now preserves structured term and unit information, and the DataMap clipboard handling was aligned with the shared clipboard implementation.
* *Electron - Main:** Fix DataMap deletion so the persisted parent ARC is updated before the DataMap itself is removed.
* **Tests:** Regression coverage was extended for annotation-table header detection, DataMap cell positioning, plain TSV pasting, structured terms and units, clipboard fallbacks, and malformed structured clipboard data.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

- [x] Components
- [x] Client
- [x] Main

# How Has This Been Tested?

- [x] **Paste target:** Verifies that pasting uses the clicked column when an older selection exists elsewhere.
- [x] **DataMap column alignment:** Verifies that copied terms remain in their grid columns and do not shift following values.
- [x] **DataMap plain-text paste:** Verifies that ordinary TSV values remain separate cells and are not interpreted as ontology metadata.
- [x] **DataMap structured paste:** Verifies that structured terms and units retain their ontology metadata when pasted.
- [x] **Clipboard codec:** Verifies round-tripping of all supported composite cell types and rejects malformed or unsupported payloads.
- [x] **HTML clipboard fallback:** Verifies that the HTML fallback preserves structured cell metadata and ignores unrelated HTML.
- [x] **Clipboard index wrapping:** Verifies repeated clipboard ranges wrap correctly and invalid empty ranges are rejected.
- [x] **Delete Datamap and recreate:** Verifies that the deleted Datamap is not recreated when a new Datamap is added in the same session

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

> The Zen of GitHub.

- Practicality beats purity
- Speak like a human
- Encourage flow
- Favor focus over features
- Design for failure
- Keep it logically awesome

src/Client/Spreadsheet/Controller/Clipboard.fs
```diff
@@ -4,60 +4,62 @@ open Fable.Core
...
 let pasteCellsByIndexExtend (index: CellCoordinate) (state: Spreadsheet.Model) : JS.Promise<Spreadsheet.Model> = promise {
-    let! tab = navigator.clipboard.readText ()
+    let! content = ClipboardCodec.read ()
     let header = Generic.getHeader index.x state
-    let cells = CompositeCell.fromTabTxt tab header
+
+    let cells =
+        content.Payload
+        |> Option.map (fun payload ->
+            payload.Rows
```
caroott: `Rows` is a 2-D payload, but this keeps only the first cell from each row as far as i can tell. A 2x2 payload like `[[A; B]; [C; D]]` becomes `[A; C]`, so `B` and `D` are lost? The annotation-table copy path can write multi-column selections as a matrix i think.

src/Client/Spreadsheet/Controller/Clipboard.fs
```diff
@@ -4,60 +4,62 @@ open Fable.Core
...
-    let tab = cell.ToTabStr()
-    navigator.clipboard.writeText (tab)
+let private getCellsByIndex (indices: CellCoordinate[]) (state: Spreadsheet.Model) =
+    indices |> Array.map (fun index -> Generic.getCell (index.x, index.y) state)
 
-let copyCells (cells: CompositeCell[]) : JS.Promise<unit> =
-    let tab = CompositeCell.ToTabTxt cells
-    navigator.clipboard.writeText (tab)
+let private writeCells plainText (cells: CompositeCell[]) =
+    let rows = cells |> Array.map Array.singleton
```
caroott: This turns every cell into its own row, did you check if it retains the original selection shape? Four cells from a 2x2 selection are encoded as `[[A]; [B]; [C]; [D]]` instead of `[[A; B]; [C; D]]`. `copyCellsByIndex` still has the coordinates before `getCellsByIndex` drops them. Can we use those coordinates to build the matrix, or enforce a single-column invariant before reaching this helper?

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,309 @@
...
+    ]
+    |> unbox<Payload>
+
+let encode payload = JS.JSON.stringify payload
+
+let tryDecode json =
+    try
+        let payload = JS.JSON.parse (json) |> unbox<Payload>
+
+        if
```
caroott, line 189: An empty row passes this validation because `Array.forall` returns `true` for an empty array. That makes `Rows = [[A]; []; [B]]` a valid structured payload. The annotation-table paste path rejects an empty row later, while `pasteCellsByIndexExtend` drops it and shifts the following cells up.

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,309 @@
...
+        None
+
+let private escapeHtml (text: string) =
+    text.Replace("&", "&amp;").Replace("<", "&lt;").Replace(">", "&gt;").Replace("\"", "&quot;").Replace("'", "&#39;")
+
+let createHtmlRepresentation (plainText: string) (cells: CompositeCell[][]) =
+    let encodedPayload = cells |> createPayload |> encode |> JS.encodeURIComponent
+
+    let tableRows =
+        plainText.TrimEnd([| '\r'; '\n' |]).Split([| "\r\n"; "\n"; "\r" |], System.StringSplitOptions.None)
```
caroott: When `A\n` represents a two-row range whose last row is empty, `TrimEnd` turns it into the same HTML table as `A`. The final empty row is therefore missing from the HTML representation. Since `cells` already contains the matrix shape, can the HTML rows be built from that matrix instead of trimming the TSV first?

src/Client/Spreadsheet/Controller/Clipboard.fs
Etschbeijer: Fixed in commit: ad23be0ffb2d01ecd3ef962df776ebcbe3b4b7bc

src/Client/Spreadsheet/Controller/Clipboard.fs
Etschbeijer: Fixed in commit: ad23be0ffb2d01ecd3ef962df776ebcbe3b4b7bc

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: ad23be0ffb2d01ecd3ef962df776ebcbe3b4b7bc

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: ad23be0ffb2d01ecd3ef962df776ebcbe3b4b7bc

src/Components/src/Composite/DataMapTable/Types.fs
```diff
@@ -28,7 +28,48 @@ module ARCtrlExtensions =
...
+                        let source = Swate.Components.ClipboardCodec.toCompositeCell dto
+                        let target = this.GetCell(columnIndex, rowIndex)
+
+                        let cell =
+                            match source, columnIndex with
+                            | CompositeCell.Unitized(_, unit), DataMapIndices.Unit -> CompositeCell.createTerm unit
+                            | CompositeCell.Term term, _ when this.GetHeader(columnIndex).IsTermColumn ->
+                                CompositeCell.createTerm term
+                            | CompositeCell.Data _, DataMapIndices.Data -> source
+                            | _ -> target.UpdateMainField(source.ToString())
```
caroott, line 68: Pasting FreeText "foo" into a Unit cell containing metre changes its name to foo while retaining metre’s ontology accession. This reuses a text-edit helper where a source-to-destination conversion is needed. Construct a fresh term when converting structured text into a term cell.

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,315 @@
...
+type Payload =
+    abstract Version: int
+    abstract Rows: CellDto[][]
+
+type ClipboardContent = {
+    PlainText: string
+    Payload: Payload option
+}
+
+type private CellData = {
```
caroott, line 38: This duplicates CellDto as CellData, then copies fields through string-keyed createObj and unbox. A typed  DTO could remove duplication and unchecked field-name mapping. The write path also serializes the same payload twice.

caroott (commented):
Both `Spreadsheet.Controller.Clipboard.pastePayloadByIndexExtend` and `DataMap.PastePayload` traverse the clipboard matrix, calculate target coordinates, convert cells, and write them. Their behavior differs when targeting a DataMap:

| Operation | Spreadsheet controller | DataMap component |
|---|---|---|
| Paste `4 metre` into Unit | Keeps `Unitized`; the setter requires `Term`, so it would throw | Extracts the `metre` term |
| Paste text into an existing term | Creates a fresh term | Retains the destination’s old ontology ID |
| Paste beyond existing rows | Does not grow the DataMap | Adds rows first |

Maybe remove the controller path if unused. Otherwise, delegate its DataMap branch to one shared implementation, while retaining separate annotation-table rules where the destination models differ.

src/Components/src/Composite/DataMapTable/Types.fs
Etschbeijer: I would adapt this in a new PR because it is a changed to the logic that was implemented so far.

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: 455daf989f112eedd05f1ad40ae01765b9532e84

Etschbeijer:
> Both `Spreadsheet.Controller.Clipboard.pastePayloadByIndexExtend` and `DataMap.PastePayload` traverse the clipboard matrix, calculate target coordinates, convert cells, and write them. Their behavior differs when targeting a DataMap:
> Operation 	Spreadsheet controller 	DataMap component
> Paste `4 metre` into Unit 	Keeps `Unitized`; the setter requires `Term`, so it would throw 	Extracts the `metre` term
> Paste text into an existing term 	Creates a fresh term 	Retains the destination’s old ontology ID
> Paste beyond existing rows 	Does not grow the DataMap 	Adds rows first
> 
> Maybe remove the controller path if unused. Otherwise, delegate its DataMap branch to one shared implementation, while retaining separate annotation-table rules where the destination models differ.

Fixed in commit: 455daf989f112eedd05f1ad40ae01765b9532e84

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,284 @@
...
+    text.Replace("&", "&amp;").Replace("<", "&lt;").Replace(">", "&gt;").Replace("\"", "&quot;").Replace("'", "&#39;")
+
+let private createHtmlRepresentationFromPayload (cells: CompositeCell[][]) (payloadText: string) =
+    let encodedPayload = payloadText |> JS.encodeURIComponent
+
+    let tableRows =
+        cells
+        |> Array.map (fun row ->
+            row
+            |> Array.map (fun cell -> $"<td>{cell.ToClipboardStr() |> escapeHtml}</td>")
```
caroott, line 177: This breaks pasting into Excel, Google Sheets and LibreOffice. All three prefer text/html over text/plain when both are on the clipboard, and this is the first time we write text/html at all. ToClipboardStr() for a Term is "name\tTSR\tTAN" (four fields for Unitized), and escapeHtml leaves the tabs alone, so the whole thing lands in one <td> and HTML whitespace collapsing turns it into "growth PO PO:0001" in a single cell. On main the same copy gave three cells. For multi-column selections the HTML and the plain text now disagree on column count, so the paste shape depends on the target app.

Suggest one <td> per tab separated field so the HTML grid matches the plain text grid, or build the HTML from the display value. The decoder reads the data attribute, not the cells, so either is safe.

src/Components/src/Composite/AnnotationTable/ContextMenu.fs
```diff
@@ -551,11 +590,21 @@ type AnnotationTableContextMenuUtil =
...
         ) =
         promise {
-            let! data = AnnotationTableContextMenuUtil.getCopiedCells ()
+            let! content = Swate.Components.ClipboardCodec.read ()
 
             try
                 let prediction =
-                    AnnotationTableContextMenuUtil.predictPasteBehaviour (cellIndex, arcTable, selectHandle, data)
+                    match content.Payload with
+                    | Some payload when cellIndex.y > 0 ->
```
caroott, line 598: The guard only looks at the clicked cell, not the selection. Select A0:A3 (header included), right click A2, paste: y is 2 so we take the structured path, getPasteTargetCoordinates returns the whole selection including y = 0, and pasteCells ends up calling UpdateHeader with the column's own header (predictPayloadBehaviour built the headers from targetTable itself). Net effect: a silent no-op on the header and a body-only paste, where the plain text path would have shown the AddColumns modal.

Same function also never returns PasteCases.Unknown, so an all-blank payload pastes silently instead of prompting. Probably want to fall through to the plain text path whenever any target coordinate has y <= 0, and keep the isEmpty check from predictPasteBehaviour.

src/Components/src/Composite/DataMapTable/Types.fs
```diff
@@ -28,7 +28,48 @@ module ARCtrlExtensions =
...
+                    let columnIndex = startCoordinate.x - 1 + columnOffset
+                    let rowIndex = startCoordinate.y - 1 + rowOffset
+
+                    if columnIndex < this.ColumnCount then
+                        let source = Swate.Components.ClipboardCodec.toCompositeCell dto
+                        let target = this.GetCell(columnIndex, rowIndex)
+
+                        let cell =
+                            match source, columnIndex with
+                            | CompositeCell.Unitized(_, unit), DataMapIndices.Unit -> CompositeCell.createTerm unit
```
caroott, line 64: This only rescues Unitized cells when the target is column 5. Explication (4) and ObjectType (6) are also term columns, but a Unitized source there falls through to the last arm and becomes target.UpdateMainField("5 mg"), so the term name is the literal string "5 mg" and TSR/TAN are gone. Term sources in the same columns keep their accession via the second arm, so the two cases are inconsistent.

`| CompositeCell.Unitized(_, unit), _ when this.GetHeader(columnIndex).IsTermColumn -> CompositeCell.createTerm unit`

would cover all three.

src/Components/src/Composite/DataMapTable/Types.fs
```diff
@@ -28,7 +28,48 @@ module ARCtrlExtensions =
                 |> Seq.toArray
             )
             |> Seq.toArray
-            |> CompositeCell.ToClipboardTableTxt
+
+        member this.SelectedCellsToTabText(coordinates: seq<CellCoordinate>) =
```
caroott, line 32: Switching to ToString() here is the right call for Excel and for DataMap to DataMap paste, but it changes what an annotation table receives when there is no structured payload (Firefox, or clipboard.read() denied). ARCtrl's ToString for Term is just the name, so "growth\tPO\tPO:0001" on main becomes "growth" here, and fromContentValid turns that into a Term with empty TSR/TAN. The annotation table's own copy still emits the three field form, so the two tables now speak different plain text dialects. If that's the intended trade-off it should be in the CHANGELOG, because the entry on line 32 currently claims the opposite for browsers without custom clipboard formats.

Nit: the Data arm on line 38 is redundant, ToClipboardStr and ToString both give data.Name for Data cells.

src/Client/Spreadsheet/Controller/Clipboard.fs
```diff
@@ -4,69 +4,122 @@ open Fable.Core
...
-        let emptyCell = cell.GetEmptyCellFixed()
-        Generic.setCell (index.x, index.y) emptyCell state
-        cells.Add(cell)
+    ClipboardCodec.write (CompositeCell.ToClipboardTableTxt rows) (Some rows)
+    |> Promise.start
 
-    copyCells (Array.ofSeq cells) |> Promise.start
     state
 
 let pasteCellByIndex (index: CellCoordinate) (state: Spreadsheet.Model) : JS.Promise<Spreadsheet.Model> = promise {
```
caroott, line 46: Nothing dispatches CopyCell, CutCell, PasteCell or PasteCellsExtend, so this whole module, including the DataMap +1 fix in pastePayloadByIndexExtend and the new Clipboard.tests.fs, only runs under test. The DataMap paste users actually hit is DataMapTable/Helper.pasteCells. Worth stating in the PR description so nobody assumes the "+1" bug fix lands via this path.

While it's here: pasteCellByIndex in the DataMap view still throws for a Unitized payload into the Unit column. ConvertToValidCell keeps it Unitized (the header is a term column), then DataMap.SetCell calls AsTerm and fails with "Not a Term Cell." PastePayload handles exactly this case, so the two entry points disagree. Pre-existing for the plain text branch, but new for the payload branch.

src/Client/Spreadsheet/Controller/Clipboard.fs
```diff
@@ -4,69 +4,122 @@ open Fable.Core
...
+    let rows = getCellRowsByIndex indices state
+    ClipboardCodec.write (CompositeCell.ToClipboardTableTxt rows) (Some rows)
 
 let cutCellByIndex (index: CellCoordinate) (state: Spreadsheet.Model) : Spreadsheet.Model =
     let cell = Generic.getCell (index.x, index.y) state
-    // Remove selected cell value
-    let emptyCell = cell.GetEmptyCellFixed()
-    Generic.setCell (index.x, index.y) emptyCell state
-    copyCell cell |> Promise.start
+    Table.clearCells [| index |] state |> ignore
```
caroott, line 30: Order matters here for DataMap Data cells. Generic.getCell on DataMap column 0 wraps the live DataContext, clearCells then blanks that same object, and only afterwards do we serialise cell.ToClipboardStr() and the payload. So cutting a DataMap Data cell puts an empty string and an empty Data DTO on the clipboard. Same on line 39. Copy before clearing, or snapshot with cell.Copy(). Also, if all three write fallbacks reject, the rejection goes into Promise.start unhandled after the cells were already cleared, which is silent data loss. Copy first, clear on success, would fix both.

src/Electron/src/Main/ARCtrlExtensions/DataMap.fs
```diff
@@ -41,14 +58,33 @@ module DataMapExtensions =
...
+                    | None ->
+                        let parentPath = DatamapParentInfo.toFolderPath parentInfo
+
+                        return
+                            Error(
+                                exn
+                                    $"The DataMap could not be deleted because parent '{parentPath}' was not found on disk. Refresh the File Explorer and try again."
+                            )
+                    | Some parentUpdateContract ->
+                        match! fullFillContractBatchAsync arcPath [| parentUpdateContract; deleteContract |] with
```
caroott, line 79: fullFillContractBatchAsync runs every contract regardless of earlier failures (CrossAsync.startSequential in ARCtrl only folds the errors at the end). So if writing the parent fails, say the assay xlsx is locked, the delete contract still removes isa.datamap.xlsx, we return Error, and TrySetDataMap never runs. In-memory still has a DataMap, disk has a parent that still embeds one, and the standalone file is gone. Probably fine to accept, but it's not the clean rollback the error message implies. Running the parent update on its own first and only issuing the delete on Ok would keep the two states aligned.

Separately, the fix itself is right: ArcAssay.toFsWorkbook embeds the DataMap as a worksheet in the parent file and the loader prefers that copy, which is why main's file-only delete resurrected it.

src/Electron/src/Main/ARCtrlExtensions/DataMap.fs
```diff
@@ -41,14 +58,33 @@ module DataMapExtensions =
             | Some dataMap ->
                 let deleteContract = dataMap.ToDeleteContract(parentInfo)
 
-                match! fullFillContractBatchAsync arcPath [| deleteContract |] with
+                match! ARC.LoadAsyncSwate arcPath with
```
caroott, line 61: LoadAsyncSwate walks the entire ARC tree and reads every ISA file just to get one parent object. On main a DataMap delete touched one file; now it's O(size of ARC) plus a full re-serialisation of the parent workbook. There is no lighter loader in LoadAsync.fs today, so this may be the pragmatic choice, but for ARCs with dozens of assays it's a noticeable regression. A targeted read of the parent's own xlsx via its read contract would do.

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,284 @@
...
+        let! items = clipboard.read ()
+
+        let item =
+            items
+            |> Array.tryFind (fun item -> hasType "text/plain" item && hasType MimeType item)
+            |> Option.orElseWith (fun () ->
+                items
+                |> Array.tryFind (fun item -> hasType "text/plain" item && hasType "text/html" item)
+            )
+            |> Option.orElseWith (fun () -> items |> Array.tryFind (hasType "text/plain"))
```
caroott, line 257: If an item carries our MIME type but no text/plain, none of the three predicates match and we fall through to readText(), dropping the payload. Unlikely from our own writer, but the second fallback also picks any text/plain + text/html item, which need not be the one holding the payload when several items are present. Cheaper to pick the first item that has MimeType, then the first with text/html, then text/plain, and read text/plain from it if present.

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,284 @@
...
+        data.SelectorFormat <-
+            cell.SelectorFormat
+            |> Option.ofObj
+            |> Option.filter (System.String.IsNullOrWhiteSpace >> not)
+
+        CompositeCell.createData data
+    | "freetext" -> CompositeCell.createFreeText cell.Value
+    | kind -> failwith $"Unknown clipboard cell kind: {kind}"
+
+let validateCellDto (cell: CellDto) =
```
caroott, line 109: validateCellDto checks the fields exist but not that they are strings. {"Kind":"freetext","Value":5} passes and produces CompositeCell.FreeText 5 with a JS number inside, which then flows into ToTermCell and String.concat. A typeof check per field would close this. The rest of tryDecode holds up: a string Version is rejected, Rows as an object throws inside forall and is caught, unknown Kind is rejected.

src/Components/src/Composite/AnnotationTable/ContextMenu.fs
```diff
@@ -364,6 +360,49 @@ type AnnotationTableContextMenuUtil =
                     coordinates = groupedCellCoordinates
                 |}
 
+    static member predictPayloadBehaviour
```
caroott, line 363: No test calls this. The PR headline is structured terms and units surviving copy/paste, and the DataMap side has tests for that, but the annotation table side has none. Nor is tryPasteCopiedCells tested end to end, so the y > 0 guard on line 598 and the header-row case above are unexercised. One test with a Term payload and a Unitized payload into an ArcTable column asserting TSR/TAN would cover the main claim.

Nit: the failwith on line 379 is unreachable, tryDecode already rejects empty rows.

src/Components/src/ARCtrl/ClipboardCodec.fs
```diff
@@ -0,0 +1,284 @@
...
+            None
+
+let private hasType mimeType (item: ClipboardItem) = item.types |> Array.contains mimeType
+
+let private tryReadPayload (item: ClipboardItem) = promise {
+    try
+        if hasType MimeType item then
+            let! blob = item.getType MimeType
+            let! json = blob.text ()
+            return tryDecode json
```
caroott, line 210: The write fallback chain (typed item, then HTML only item, then writeText) is what commit 8099e0a1 says fixed copy outside Swate, but nothing tests it. Clipboard.tests.fs only mocks read() returning an empty array, which real browsers never do; a real "no custom format" clipboard returns an item with text/plain and text/html types. A mock where write throws once, and one where read() returns such an item, would cover the actual fallback scenarios.

src/Components/src/JsBindings/Clipboard.fs
```diff
@@ -0,0 +1,59 @@
...
+[<AutoOpen>]
+module GlobalBindings =
+
+    [<Emit("navigator")>]
+    let navigator: Navigator = jsNative
+
+module ClipboardBindings =
+
+    [<Emit("typeof DOMParser !== 'undefined'")>]
+    let isHtmlParserAvailable: bool = jsNative
```
caroott, line 35: isHtmlParserAvailable is only referenced from tests, and the same goes for createHtmlRepresentation in ClipboardCodec.fs and getCopiedCells in ContextMenu.fs (line 178, no callers left). Drop them or move the test-only ones into the test project.

caroott (dismissed):

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: 83a9c22

src/Components/src/Composite/AnnotationTable/ContextMenu.fs
Etschbeijer: Fixed in commit: 83a9c22 and 4965bd6

src/Components/src/Composite/DataMapTable/Types.fs
Etschbeijer: Fixed in commit: 83a9c22

src/Client/Spreadsheet/Controller/Clipboard.fs
Etschbeijer: Fixed in commit: 83a9c22

src/Client/Spreadsheet/Controller/Clipboard.fs
Etschbeijer: Fixed in commits: 83a9c22 & 4965bd6 & 7959d9e

src/Electron/src/Main/ARCtrlExtensions/DataMap.fs
Etschbeijer: Fixed in commit: 4965bd6

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: 83a9c22

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: 83a9c22

src/Components/src/Composite/AnnotationTable/ContextMenu.fs
Etschbeijer: Fixed in commit: 4965bd6

src/Components/src/ARCtrl/ClipboardCodec.fs
Etschbeijer: Fixed in commit: 4965bd6

src/Components/src/Composite/DataMapTable/Types.fs
Etschbeijer: This is intented behaviour. I will expand the changelog and upadte the netry on line 32.

src/Components/src/JsBindings/Clipboard.fs
Etschbeijer: Removed in commit: 83a9c22 & 397c9e39de6ab3a3c2929bb45c6b480d65ef873f

src/Electron/src/Main/ARCtrlExtensions/DataMap.fs
Etschbeijer: Fixed in commit: c604386

src/Components/src/Composite/DataMapTable/Types.fs
Etschbeijer: Fixed in commit: 83a9c22

Freymaurer (approved):

Issue #1279, opened 2026-08-04
[Feature Request]: Add validation package from AVPR


Brilator:
### Is your feature request related to a problem?

SwatApp should provide the option to add validation packages, just as ARCitect did.

<img width="1607" height="663" alt="Image" src="https://github.com/user-attachments/assets/1d6f89f4-92f2-4607-87c7-cede1a27ceaa" />

Even though the CQC is run on the DataHUB, adding the validation should be user-friendly and close to ARC creation.

Etschbeijer:
Released in version 2.3.0

PR #1338, opened 2026-09-08
Wire up ValidationPackageBrowser with PageState

Freymaurer:
- Move `ValidationPackageSelector` from Composite to Page components
- Rename `ValidationPackageSelector` to `ValidationPackageBrowser`
- Wire up `ValidationPackageBrowser` with avpr api (https://avpr.nfdi4plants.org/swagger/index.html), `api/v1/packages`. This is called from main
- Add sidebar button to open up PageState for `ValidationPackageBrowser`. This will check if config exists in `.arc/validation_packages.yml`, if not uses empty class. If any changes are done then config is written to the same path. Uses ARCtrl path literals, as well as ARCtrl yaml parsing.
- Uses generic file read and file write IPC calls, to not bloat ipc layer with a hyper-specific writeValidationPackagesYml


src/Electron/src/Renderer/Components/MainContent/ValidationPackageBrowserTarget.fs
```diff
@@ -0,0 +1,117 @@
...
+open Swate.Components.Page.ValidationPackageBrowser.Helper
+open Swate.Components.Page.ValidationPackageBrowser.Types
+open Swate.Components.Primitive.ErrorModal.Context
+open Swate.Electron.Shared.FileIOHelper
+open Swate.Electron.Shared.FileIOTypes
+open Renderer.Components.Helper.ArcVaultHelper
+
+module private ValidationPackageBrowserHelper =
+
+    let configFolderPath = ARCtrl.ArcPathHelper.ARCConfigFolderName
```
Etschbeijer: I think we can skip this binding

src/Electron/src/Renderer/Components/MainContent/ValidationPackageBrowserTarget.fs
```diff
@@ -0,0 +1,117 @@
...
+
+    let configFileName = ARCtrl.ArcPathHelper.ValidationPackagesYamlFileName
```
Etschbeijer: I think we can skip this binding

src/Electron/src/Renderer/Components/MainContent/ValidationPackageBrowserTarget.fs
```diff
@@ -0,0 +1,117 @@
...
+
+[<ReactComponent(true)>]
+let ValidationPackageBrowserTarget () =
+    let errorModal = useErrorModalCtx ()
+    let appStateCtx = Renderer.Context.AppStateContext.useAppStateCtx ()
+
+    let onError =
+        createErrorModalCallback errorModal.enqueue "Validation packages" appStateCtx
+
+    let config, setConfig =
```
Etschbeijer, line 117: I think we should distinguish between “missing config” and “loading/failed to load” here. 

I think that at the moment, a read/parse failure would leave emptyConfig() active. As a resulst, the browser would still become editable and a later submit may overwrite an existing validation package file, which was never successfully loaded in the first place.

I think that disabling editing and submitting until loading is resolved succesfully would prevent this bug. We could treat "OK false" as the intentional empty-config case.

What do you think? Am I overlooking something?

src/Components/src/Page/ValidationPackageBrowser/Helper.fs
```diff
@@ -248,6 +248,26 @@ let unlistedNames (config: ValidationPackagesConfig) (packages: ValidationPackag
...
+    |> Array.map (fun (_, versions) ->
+        versions
+        |> Array.reduce (fun current candidate ->
+            let currentSemVer =
+                ARCtrl.Helper.SemVer.SemVer.tryOfString (toVersionString current)
+
+            let candidateSemVer =
+                ARCtrl.Helper.SemVer.SemVer.tryOfString (toVersionString candidate)
+
+            match currentSemVer, candidateSemVer with
```
Etschbeijer: I think this becomes order-dependent when one version is an invalid SemVer.

When you have the following constellation, "invalid; valid", it will be invalid but when it is [valid; invalid] it will become valid or not?

We should also check for the "None-Some" constellations here, that might help and add a reverse order test.

Etschbeijer (changes requested):

src/Electron/src/Renderer/Components/MainContent/ValidationPackageBrowserTarget.fs
Freymaurer: I added a small state to track error states internally. The error is shown in a minimal daisui alert with the information, that dismissing the error might lead to loss of existing validation package configuration.

As long as the error exists and is not dismissed, we do not show the vp-browser.

src/Components/src/Page/ValidationPackageBrowser/Helper.fs
Freymaurer: This whole parsing thing was a dumb idea, i totally overlooked it. I swapped it out for a type transformation from `ValidationPackageDTO` to `SemVer`. Way faster, no error possibility.

Etschbeijer:
Closes #1279

Etschbeijer (approved):

Issue #1331, opened 2026-08-25
[Bug]: Opening a folder that is not an ARC crashes the app with an uncaught "spawn git EMFILE" exception instead of a clear error


Gitarackur:
### What happened?

Opening a folder that is not an ARC in the Swate desktop app (in this case, a local checkout of the Swate source project folder) crashes the app with an uncaught JavaScript exception in the main process (`spawn git EMFILE`) instead of showing a clear, user-friendly error that the selected folder is not an ARC. The Swate project folder was opened by mistake through Swate.

#### Steps shown in the screenshot

- Open a valid ARC in the Swate desktop app, e.g. ArcPrototype-1. The ARC opens correctly.

- Open a folder that is not an ARC.

- The app attempts to load the folder: the file tree sidebar shows the folder name and "No files found".

- An error dialog appears: "A JavaScript error occurred in the main process" with `Error: spawn git EMFILE`.

#### Observed behavior

The app does not validate that the selected folder is an ARC before attempting to load it. For a non-ARC folder that is a large git repository, git child processes are spawned until the process file-descriptor budget is exhausted, producing an uncaught exception:

```
Uncaught Exception:
Error: spawn git EMFILE
    at ChildProcess._handle.onexit (node:internal/child_process:287:19)
    at onErrorNT (node:internal/child_process:508:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

`EMFILE` is "too many open files" (errno 24). On macOS, the default per-process soft limit is 256 file descriptors (`launchctl limit maxfiles` → `256 unlimited`).

Evidence: swate-non-arc-open-emfile.png (screenshot)

#### Expected behavior

When the selected folder is not a valid ARC, Swate should detect this and show a clear, actionable error message (e.g., "The selected folder is not an ARC"), without crashing and without exhausting its file-descriptor budget when opening foreign folders.

#### Reproducibility

Reproduced by opening a local checkout of the Swate source repository as an ARC. Likely reproducible with any large non-ARC git repository.

### Host

Standalone

### Host Details

- Swate desktop app: 2.0.1 (build 2.0.1)

- OS: macOS 26.4.1 (Build 25E253)

- Architecture: arm64

- Screenshot captured: 2026-08-26

Gitarackur:
[...]

PR #1339, opened 2026-09-10
Prevent opening of none ARC folder

Etschbeijer:
* Use Electron dialgogue call for error message
* Add error message when non ARC folder is opened

### DIFF
--- a/src/Electron/src/Main/ArcVault/ArcVault.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVault.fs
@@ -79,6 +79,8 @@ type ArcVault(window: BrowserWindow) =
         this.arc <- Some arc
         this.window.title <- arc.Identifier
 
+    member this.ClearArc() = this.arc <- None
+
     /// Sets the dirty marker for unsaved in-memory ARC mutations.
     member this.RefreshHasUnsavedArcChangesFlag() =
         // Use this value to only send updates to the renderer when the dirty state actually changes. This avoids redundant updates.
@@ -407,8 +409,19 @@ module ArcVaultExtensions =
 
                 swatelogfn this.window.id "path: %s" normalizedPath
                 this.path <- Some normalizedPath
-                do! this.Startup()
-                sendMsg.pathChange (Some normalizedPath)
+
+                try
+                    do! this.Startup()
+                    sendMsg.pathChange (Some normalizedPath)
+                with error ->
+                    // Loading happens after the path has been assigned so the watcher and ARCtrl can use it.
+                    // Restore the empty-vault state when loading fails; otherwise a non-ARC folder leaves this
+                    // window permanently bound to a path that was never opened successfully.
+                    do! this.StopFileWatcher()
+                    this.path <- None
+                    this.ClearArc()
+                    this.fileTree.Clear()
+                    return raise error
         }
 
         member this.CreateARC(path: string, identifier: string) = promise {
--- a/src/Electron/src/Main/ArcVault/ArcVault.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVault.fs
@@ -391,8 +391,8 @@ module ArcVaultExtensions =
 
         /// This functions should be called once, when an vault is first started with a path
         member this.Startup() = promise {
-            this.StartFileWatcher()
             do! this.LoadArc()
+            this.StartFileWatcher()
             this.window.title <- this.arc.Value.Identifier
         }
 
@@ -414,9 +414,8 @@ module ArcVaultExtensions =
                     do! this.Startup()
                     sendMsg.pathChange (Some normalizedPath)
                 with error ->
-                    // Loading happens after the path has been assigned so the watcher and ARCtrl can use it.
-                    // Restore the empty-vault state when loading fails; otherwise a non-ARC folder leaves this
-                    // window permanently bound to a path that was never opened successfully.
+                    // The path is assigned before loading so ARCtrl can use it. Restore the empty-vault state
+                    // when startup fails; otherwise the window remains bound to a path that was never opened.
                     do! this.StopFileWatcher()
                     this.path <- None
                     this.ClearArc()
@@ -673,14 +672,26 @@ type ArcVaults() =
         let id = window.id
         let vault = ArcVault(window)
         this.Vaults.Add(id, vault)
-        do! vault.OpenARC(path)
 
-        this.OnCloseWindow(window, vault, id)
+        try
+            do! vault.OpenARC(path)
 
-        window.focus ()
-        swatelogfn id "Register window"
+            this.OnCloseWindow(window, vault, id)
 
-        return id
+            window.focus ()
+            swatelogfn id "Register window"
+
+            return id
+        with error ->
+            // The normal close lifecycle is registered only for successfully initialized vaults.
+            // Clean up directly so a failed open cannot leave an empty window or orphaned entry.
+            do! vault.StopFileWatcher()
+            this.Vaults.Remove(id) |> ignore
+
+            if not (window.isDestroyed ()) then
+                window.destroy ()
+
+            return raise error
     }
 
     member this.RegisterVaultWithNewArc(path: string, newIdentifier: string) : Fable.Core.JS.Promise<int> = promise {
--- a/src/Electron/src/Main/ArcVault/ArcVault.fs
+++ b/src/Electron/src/Main/ArcVault/ArcVault.fs
@@ -390,9 +390,13 @@ module ArcVaultExtensions =
         }
 
         /// This functions should be called once, when an vault is first started with a path
-        member this.Startup() = promise {
+        member this.Startup(?startFileWatcher: unit -> unit) = promise {
             do! this.LoadArc()
-            this.StartFileWatcher()
+
+            match startFileWatcher with
+            | Some start -> start ()
+            | None -> this.StartFileWatcher()
+
             this.window.title <- this.arc.Value.Identifier
         }
 
--- a/src/Electron/src/Main/IPC/IArcVaultsApi.fs
+++ b/src/Electron/src/Main/IPC/IArcVaultsApi.fs
@@ -120,7 +120,7 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
 let private showArcOpenError (window: BaseWindow option) (arcPath: string) (error: exn) =
     let options =
         Dialog.ShowMessageBoxSync.Options(
-            "The selected folder is not a valid ARC and could not be opened.",
+            "The ARC could not be opened.",
             ``type`` = Enums.Dialog.ShowMessageBoxSync.Options.Type.Error,
             title = "Could not open ARC",
             detail = $"Folder: {arcPath}\n\n{error.Message}"
--- a/src/Electron/src/Main/IPC/IArcVaultsApi.fs
+++ b/src/Electron/src/Main/IPC/IArcVaultsApi.fs
@@ -117,6 +117,65 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
         |> fun rendererApi -> rendererApi.gitRepositoryInitialized arcPath
     )
 
+let private showArcOpenError (window: BaseWindow option) (arcPath: string) (error: exn) =
+    let options =
+        Dialog.ShowMessageBoxSync.Options(
+            "The selected folder is not a valid ARC and could not be opened.",
+            ``type`` = Enums.Dialog.ShowMessageBoxSync.Options.Type.Error,
+            title = "Could not open ARC",
+            detail = $"Folder: {arcPath}\n\n{error.Message}"
+        )
+
+    dialog.showMessageBoxSync (?window = window, options = options) |> ignore
+
+let private validateArcRoot (arcPath: string) = promise {
+    let investigationPath =
+        ARCtrl.ArcPathHelper.combine arcPath ARCtrl.ArcPathHelper.InvestigationFileName
+
+    let! investigationExists = pathExistsAsync investigationPath
+
+    if investigationExists then
+        return Ok()
+    else
+        return
+            Error(
+                exn
+                    $"The folder does not contain the required ARC investigation file '{ARCtrl.ArcPathHelper.InvestigationFileName}'."
+            )
+}
+
+let private openArcAtPath (event: IpcMainInvokeEvent) (requestedPath: string) = promise {
+    let window = dialogParentFromIpcEvent event
+
+    let reportError arcPath error =
+        showArcOpenError window arcPath error
+        Error error
+
+    let normalizedPathResult =
+        try
+            Ok(PathHelpers.normalizePath requestedPath)
+        with error ->
+            Error error
+
+    match normalizedPathResult with
+    | Error error -> return reportError requestedPath error
+    | Ok arcPath ->
+        try
+            let! arcPathExists = pathExistsAsync arcPath
+
+            if not arcPathExists then
+                return reportError arcPath (exn $"The ARC cannot be found at location: '{arcPath}'.")
+            else
+                match! validateArcRoot arcPath with
+                | Error error -> return reportError arcPath error
+                | Ok() ->
+                    let windowId = windowIdFromIpcEvent event
+                    let! disposition = ARC_VAULTS.OpenOrFocusArc(windowId, arcPath)
+                    return Ok disposition
+        with error ->
+            return reportError arcPath error
+}
+
 /// This depends on the types in this file, but the types on this file must call this to bind IPC calls :/
 let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
     openARC =
@@ -136,26 +195,15 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
             elif r.filePaths.Length <> 1 then
                 return Error(exn "Not exactly one path")
             else
-                let arcPath = r.filePaths |> Array.exactlyOne |> PathHelpers.normalizePath
-
-                let windowId = windowIdFromIpcEvent event
-                let! disposition = ARC_VAULTS.OpenOrFocusArc(windowId, arcPath)
-                return Ok(Some(ArcOpenDisposition.path disposition))
+                match! openArcAtPath event (Array.exactlyOne r.filePaths) with
+                | Ok disposition -> return Ok(Some(ArcOpenDisposition.path disposition))
+                | Error error -> return Error error
         }
     openARCByPath =
         fun (arcPath: string) -> promise {
-            try
-                let arcPath = PathHelpers.normalizePath arcPath
-                let! arcPathExists = pathExistsAsync arcPath
-
-                if not arcPathExists then
-                    return Error(exn $"The ARC cannot be found at location: '{arcPath}'.")
-                else
-                    let windowId = windowIdFromIpcEvent event
-                    let! disposition = ARC_VAULTS.OpenOrFocusArc(windowId, arcPath)
-                    return Ok(ArcOpenDisposition.path disposition)
-            with e ->
-                return Error e
+            match! openArcAtPath event arcPath with
+            | Ok disposition -> return Ok(ArcOpenDisposition.path disposition)
+            | Error error -> return Error error
         }
     createARC =
         fun (request: CreateArcRequest) -> promise {
--- a/tests/Electron.Core/ArcVaultHelper.test.fs
+++ b/tests/Electron.Core/ArcVaultHelper.test.fs
@@ -630,4 +630,29 @@ Vitest.describe (
                         Vitest.expect(loadedArc.GetRun("Run With DataMap").DataMap.Value.StaticHash).not.toBe (0)
                     })
         )
+
+        Vitest.test (
+            "OpenARC restores an empty vault after opening a non-ARC folder fails",
+            fun () -> promise {
+                let! folderPath = TestHelpers.createTempDirectoryAsync "swate-open-invalid-arc-"
+
+                try
+                    let vault = ArcVault(TestHelpers.testWindow ())
+                    let mutable failed = false
+
+                    try
+                        do! vault.OpenARC folderPath
+                    with _ ->
+                        failed <- true
+
+                    Vitest.expect(failed).toBe (true)
+                    Vitest.expect(vault.path).toEqual (None)
+                    Vitest.expect(vault.arc).toEqual (None)
+                    Vitest.expect(vault.watcher).toEqual (None)
+                    do! TestHelpers.removeDirectoryAsync folderPath
+                with error ->
+                    do! TestHelpers.removeDirectoryAsync folderPath
+                    return raise error
+            }
+        )
 )
--- a/tests/Electron.Core/ArcVaultHelper.test.fs
+++ b/tests/Electron.Core/ArcVaultHelper.test.fs
@@ -655,4 +655,27 @@ Vitest.describe (
                     return raise error
             }
         )
+
+        Vitest.test (
+            "Startup does not start a watcher when ARC loading fails",
+            fun () -> promise {
+                let! folderPath = TestHelpers.createTempDirectoryAsync "swate-invalid-arc-startup-"
+
+                try
+                    let vault = ArcVault(TestHelpers.testWindow ())
+                    vault.path <- Some folderPath
+                    let mutable watcherStartCount = 0
+
+                    try
+                        do! vault.Startup(startFileWatcher = (fun () -> watcherStartCount <- watcherStartCount + 1))
+                    with _ ->
+                        ()
+
+                    Vitest.expect(watcherStartCount).toBe (0)
+                    do! TestHelpers.removeDirectoryAsync folderPath
+                with error ->
+                    do! TestHelpers.removeDirectoryAsync folderPath
+                    return raise error
+            }
+        )
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🐛 Fixed
 
+-   Show a native error message instead of crashing or silently failing when a selected folder is not a valid ARC.
 -   Keep the ARC selector at a stable width and truncate long ARC names, and close its dropdown only after direct or overflow actions are invoked without dismissing the overflow menu prematurely.
 
 ## 2.1.0 - 2026-08-19
--- a/src/Electron/src/Renderer/Components/Helper/ArcVaultHelper.fs
+++ b/src/Electron/src/Renderer/Components/Helper/ArcVaultHelper.fs
@@ -44,9 +44,8 @@ let ensureNotesFolderIfEnabled (onError: string -> unit) : JS.Promise<unit> = pr
 
 let openArc (onError: string -> unit) : JS.Promise<bool> = promise {
     match! Api.ipcArcVaultApi.openARC () with
-    | Error exn ->
-        onError exn.Message
-        return false
+    // ARC-open failures are presented by the Electron main process in a native dialog.
+    | Error _ -> return false
     | Ok None -> return false
     | Ok(Some _) ->
         do! ensureNotesFolderIfEnabled onError
@@ -55,9 +54,8 @@ let openArc (onError: string -> unit) : JS.Promise<bool> = promise {
 
 let openArcByPath (onError: string -> unit) (arcPath: string) : JS.Promise<bool> = promise {
     match! Api.ipcArcVaultApi.openARCByPath arcPath with
-    | Error exn ->
-        onError exn.Message
-        return false
+    // ARC-open failures are presented by the Electron main process in a native dialog.
+    | Error _ -> return false
     | Ok _ ->
         do! ensureNotesFolderIfEnabled onError
         return true


Etschbeijer:
grafik

Etschbeijer:
# Description

* **ARC opening:** Validate selected folders before opening them as ARCs and show a native Electron error dialog when the selected folder is not a valid ARC.
* **ARC lifecycle:** Restore the vault to an empty state when opening an ARC fails, so the window is not left bound to an invalid path.
* **File watcher:** Load the ARC before starting the file watcher, preventing non-ARC folders from being watched before validation has completed.
* **Window handling:** Clean up newly created windows and vault registrations when opening an ARC in a new window fails.
* **Renderer:** Handle ARC-open errors in the Electron main process instead of displaying an additional renderer error.
* **Tests:** Add regression coverage for restoring an empty vault after opening an invalid ARC fails.

#1331

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)

# Targeting Environment

* [x] Electron - Renderer
* [x] Electron - Main

# How Has This Been Tested?

* [x] **Invalid ARC cleanup:** Verifies that opening an invalid ARC fails and restores the vault path, ARC state, and file watcher to their empty state.
* [x] **Automated tests:** Existing Electron tests pass with the updated ARC opening and cleanup behavior.

# Checklist:

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] I have added tests that prove my fix is effective or that my feature works

> The Zen of GitHub.

- Practicality beats purity
- Speak like a human
- Encourage flow
- Favor focus over features
- Design for failure
- Keep it logically awesome

caroott:
One short question: is the switch to Process Core planned soon? If yes, it comes with a different set of requirements (probably a lot less)

Etschbeijer:
> One short question: is the switch to Process Core planned soon? If yes, it comes with a different set of requirements (probably a lot less)

I think it is not planned soon at the moment. And when we change, we can simply adapt the folder checking.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -388,9 +390,13 @@ module ArcVaultExtensions =
         }
 
         /// This functions should be called once, when an vault is first started with a path
-        member this.Startup() = promise {
-            this.StartFileWatcher()
+        member this.Startup(?startFileWatcher: unit -> unit) = promise {
```
caroott, line 393: The optional `startFileWatcher` parameter only exists so the second test can count calls. The first test already checks `vault.watcher` is `None` after a failed open, and the same assertion would cover the ordering here too. I'd drop the parameter and have the test assert on `vault.watcher` instead, so production code doesn't carry a test hook.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -136,26 +195,15 @@ let api (event: IpcMainInvokeEvent) : IPCTypes.IArcVaultsApi = {
             elif r.filePaths.Length <> 1 then
                 return Error(exn "Not exactly one path")
             else
-                let arcPath = r.filePaths |> Array.exactlyOne |> PathHelpers.normalizePath
-
-                let windowId = windowIdFromIpcEvent event
-                let! disposition = ARC_VAULTS.OpenOrFocusArc(windowId, arcPath)
-                return Ok(Some(ArcOpenDisposition.path disposition))
+                match! openArcAtPath event (Array.exactlyOne r.filePaths) with
```
caroott, line 198: Two error cases here still bypass the dialog: an exception from `showOpenDialog` and the "Not exactly one path" branch. Both return `Error` and the renderer now ignores that, so they'd fail silently. Hard to hit in practice, but routing them through `showArcOpenError` keeps behaviour consistent.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -660,14 +676,26 @@ type ArcVaults() =
...
-        swatelogfn id "Register window"
+            this.OnCloseWindow(window, vault, id)
 
-        return id
+            window.focus ()
+            swatelogfn id "Register window"
+
+            return id
+        with error ->
+            // The normal close lifecycle is registered only for successfully initialized vaults.
```
caroott, line 690: `OpenARC` already stops the watcher in its own error path before rethrowing, so this `StopFileWatcher` call never has anything to do

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -79,6 +79,8 @@ type ArcVault(window: BrowserWindow) =
         this.arc <- Some arc
         this.window.title <- arc.Identifier
 
+    member this.ClearArc() = this.arc <- None
```
caroott, line 82: `SetArc` writes the window title, but `ClearArc` doesn't reset it. If `LoadArc` succeeds and something after it in `Startup` throws, the now-empty window keeps the ARC identifier as its title. Resetting to "Swate" here would cover that.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -660,14 +676,26 @@ type ArcVaults() =
...
+
+            return id
+        with error ->
+            // The normal close lifecycle is registered only for successfully initialized vaults.
+            // Clean up directly so a failed open cannot leave an empty window or orphaned entry.
+            do! vault.StopFileWatcher()
+            this.Vaults.Remove(id) |> ignore
+
+            if not (window.isDestroyed ()) then
+                window.destroy ()
```
caroott, line 696: `createWindow` awaits `loadURL`, so the renderer is already running when we destroy the window here. Any IPC it fired during startup will log "Object has been destroyed" in main. Cosmetic, and avoiding it would mean validating before creating the window, which the IPC layer already does for the common case.

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: 17f48b63a99593aafe62922272fc88acb5378821

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: 17f48b63a99593aafe62922272fc88acb5378821

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: https://github.com/nfdi4plants/Swate/commit/17f48b63a99593aafe62922272fc88acb5378821

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: https://github.com/nfdi4plants/Swate/commit/17f48b63a99593aafe62922272fc88acb5378821

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: 7ffb178c43428b4bb4d65c8ed0e7df64287d7591

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -708,49 +723,74 @@ type ArcVaults() =
...
-        let! window = createWindow ()
+        let window = createWindow ()
         let id = window.id
         let vault = ArcVault(window)
         this.Vaults.Add(id, vault)
-        do! vault.OpenARC(path)
 
-        this.OnCloseWindow(window, vault, id)
+        try
+            do! vault.OpenARC(path)
```
caroott, line 761: With `show = false` and `loadWindow` running after `OpenARC`, a new window stays invisible until the ARC is parsed and the renderer has loaded. Before, the frame appeared right away and the ARC loaded behind it. For a large ARC opened from the recent list that is a few seconds where the click looks like it did nothing. Swapping the two `do!` lines brings the frame back early, but then a failed `OpenARC` destroys a live renderer again, which the last round wanted to avoid. Since `openArcAtPath` validates before we get here, an `OpenARC` failure at this point is rare, so I'd lean towards loading first, or showing the frame early with a loading state.

tests/Electron.Core/ArcVaultHelper.test.fs
```diff
@@ -340,6 +400,88 @@ Vitest.describe (
                 Vitest.expect(aliveWindowSendCount).toBe (1)
         )
 
+        Vitest.test (
+            "RegisterVault cleans up when renderer loading fails",
```
caroott, line 404: The new tests cover the cleanup paths, but nothing exercises the fix itself. `validateArcRoot` and `showArcOpenError` are `let private` and the mock `dialog` has no `showMessageBoxSync`, so the validation could be deleted without a test going red. Same for `loadWindow`: every fixture rejects `loadFile`, so no test reaches `window.show ()`, and with `show = false` a dropped `show()` would make every window invisible. I'd add a `showMessageBoxSync` spy to the mock plus a test for a folder without `isa.investigation.xlsx`, and one success-path registration test that asserts `show` was called.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -123,45 +123,99 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
...
+let private showArcOpenError (window: BaseWindow option) (arcPath: string) (error: exn) =
+    let options =
+        Dialog.ShowMessageBoxSync.Options(
+            "The ARC could not be opened.",
+            ``type`` = Enums.Dialog.ShowMessageBoxSync.Options.Type.Error,
+            title = "Could not open ARC",
+            detail = $"Folder: {arcPath}\n\n{error.Message}"
+        )
+
+    dialog.showMessageBoxSync (?window = window, options = options) |> ignore
```
caroott, line 135: `showMessageBoxSync` blocks the main process, so every other window freezes until the user clicks OK, including in-flight saves and git operations. The rest of this file uses the async dialogs and `dialog.showMessageBox` is in the binding. Awaiting that instead keeps the other windows responsive.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -123,45 +123,99 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
...
+                        |]
+                    )
 
-                let windowId = windowIdFromIpcEvent event
-                let! disposition = ARC_VAULTS.OpenOrFocusArc(windowId, arcPath)
-                return Ok(Some(ArcOpenDisposition.path disposition))
+                if r.canceled then
+                    return Ok None
+                elif r.filePaths.Length <> 1 then
+                    let error = exn "Not exactly one path"
```
caroott, line 203: The multi-selection case renders `Folder: a, b` and the outer handler renders a bare `Folder:` line. Making the folder line optional in `showArcOpenError` would keep the dialog tidy in those two cases.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -123,45 +123,99 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
...
+        Dialog.ShowMessageBoxSync.Options(
+            "The ARC could not be opened.",
+            ``type`` = Enums.Dialog.ShowMessageBoxSync.Options.Type.Error,
+            title = "Could not open ARC",
+            detail = $"Folder: {arcPath}\n\n{error.Message}"
+        )
+
+    dialog.showMessageBoxSync (?window = window, options = options) |> ignore
+
+let private validateArcRoot (arcPath: string) = promise {
```
caroott, line 137: Small structural note: the comment above `OpenOrFocusArc` says open decisions live in the controller, but the ARC check sits in the IPC layer, so `OpenOrFocusArc` and `OpenARCInVault` (currently without callers) still accept any folder. Fine for now since both handlers go through `openArcAtPath`, but the check would be harder to bypass inside `OpenOrFocusArc`.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -708,49 +723,74 @@ type ArcVaults() =
             this.DisposeVault(id)
         )
 
+    member private this.CleanupFailedRegistration(window: BrowserWindow, vault: ArcVault, id: int) = promise {
+        do! vault.StopFileWatcher()
+        this.Vaults.Remove(id) |> ignore
+
+        if not (window.isDestroyed ()) then
+            window.destroy ()
```
caroott, line 731: For the first window at startup this destroy means `window-all-closed` fires and the app quits without a message if `loadWindow` fails. The rejection is still dropped by the `|> ignore` in `main.fs`. Before, a blank window stayed open, which was not great either, but at least visible. A `Promise.catch` on the `RegisterVault()` call in `main.fs` that shows `dialog.showErrorBox` before quitting would make a broken build diagnosable.

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: fe83848b50142871889d9fb128b7e5489dc78785

tests/Electron.Core/ArcVaultHelper.test.fs
Etschbeijer: Fixed in commit: fe83848b50142871889d9fb128b7e5489dc78785

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: fe83848b50142871889d9fb128b7e5489dc78785

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: fe83848b50142871889d9fb128b7e5489dc78785

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commit: fe83848b50142871889d9fb128b7e5489dc78785

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: fe83848b50142871889d9fb128b7e5489dc78785

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -705,60 +732,125 @@ type ArcVaults() =
...
         this.Vaults.Add(id, vault)
-
-        do! vault.CreateARC(path, newIdentifier)
-
         this.OnCloseWindow(window, vault, id)
 
-        window.focus ()
-        swatelogfn id "Register window"
+        try
+            do! vault.CreateARC(path, newIdentifier)
```
caroott: These two lines are in the wrong order. `RegisterVaultWithValidatedArc` (line 784) calls `loadWindow` first and only then opens the ARC, and the new test "RegisterVaultWithArc shows the empty renderer before loading the ARC" pins that down. Here it is the other way round.

Since `createWindow` now passes `show = false`, nothing is on screen until `loadWindow` runs. So "create a new ARC in a new window" writes the ARC to disk and re-parses the whole thing (`CreateARC` ends in `Startup()`, which does a full `LoadArc` plus the watcher) with no window visible at all. Before this PR the window showed up immediately.

The other half: if `loadWindow` fails after `CreateARC` succeeded, `CleanupFailedRegistration` stops the watcher, drops the vault and destroys the window, but the ARC folder stays on disk. The new test at line 828 writes a real ARC into `rootPath/new-arc` and leans on its own `removeDirectoryAsync` to clean it up. Before the PR the load ran first, so a failure wrote nothing.

Swapping the two lines fixes both, and it also lets `CreateARC`'s `sendMsg.pathChange` reach a renderer that actually exists.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -423,9 +433,15 @@ module ArcVaultExtensions =
 
         /// This functions should be called once, when an vault is first started with a path
         member this.Startup() = promise {
-            this.StartFileWatcher()
             do! this.LoadArc()
-            this.window.title <- Swate.Electron.Shared.ApplicationVersion.windowTitle (Some this.arc.Value.Identifier)
+
+            if this.window.isDestroyed () then
+                return raise (exn "The ARC window was closed while the ARC was loading.")
```
caroott: This exception is correct, but it ends up in front of the user as a modal error for something they did on purpose.

For example: window B has an ARC open, the user clicks a big ARC in the recent list, window C opens and starts loading, the user changes their mind and closes C. This raises, `CleanupFailedRegistration` runs, and the error travels up through `OpenOrFocusArc` into `openArcAtPath`'s `with` branch. That calls `showArcOpenError` on the window captured at `IArcVaultsApi.fs:146`, which is window B and very much alive. B gets a blocking "The ARC could not be opened." dialog naming a folder the user just cancelled.

The test at line 655 builds exactly this scenario and checks the message, but nothing checks where the message goes.

Worth giving this one a marker (a dedicated exception type, or a `Cancelled` disposition) so `openArcAtPath` can skip the dialog for it.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -123,45 +123,84 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
...
+                    return Error error
+                else
+                    match! openArcAtPath event (Array.exactlyOne r.filePaths) with
+                    | Ok disposition -> return Ok(Some(ArcOpenDisposition.path disposition))
+                    | Error error -> return Error error
+            with error ->
+                do! showArcOpenError window None error
+                return Error error
         }
     openARCByPath =
```
caroott, line 219: The old body was wrapped in `try ... with e -> return Error e`. That wrapper is gone, and the signature still promises `JS.Promise<Result<string, exn>>`.

`openArcAtPath` covers `normalizePath` and `OpenOrFocusArc`, but `reportError` calls `showArcOpenError` outside any `try`. If `dialog.showMessageBox` throws because the parent window is gone, `openArcAtPath` rejects and the rejection goes straight out of here.

Nothing catches it downstream. `Fable.Electron.Remoting/Main.fs:127-133` hands the handler to `ipcMain.handle` untouched, so `ipcRenderer.invoke` rejects, and `match! Api.ipcArcVaultApi.openARCByPath arcPath` in the renderer throws instead of matching `Error`. Both call sites (`Navbar.fs:88`, `DataHubBrowserTarget.fs:78`) end in `Promise.start` with no catch.

tests/Electron.Core/ArcVaultHelper.wrapper.test.mts
```diff
@@ -0,0 +1,5 @@
+import { vi } from "vitest";
+
+vi.mock("fs/promises", () => import("./fs-promises.mock.mts"));
```
caroott: This line stops the `ArcVaultHelper` suite from finishing at all. I think it blocks CI.

Same machine, same compiled `output/`, four runs:

- `npx vitest run ArcVaultHelper.wrapper.test.mts`: killed at 898s, zero tests completed, `Tests no tests`.
- Same, narrowed to one trivial test with `-t "ClearArc resets the window title"`: killed at 200s, still zero.
- A one line probe containing only `import "./ArcVaultHelper.test.js";` and no `vi.mock`: 3.61s, 38 tests collected, the selected one runs. It fails on `__fsPromisesMock` being undefined, which is what you'd expect without the mock.
- Control, `output/ArcFileSystemHelper.test.js`: 19 passed in 5.54s. The environment is fine.

So the same module goes from a 3.45s import to not completing in 200 seconds, and the mock is the only difference.

I also checked whether the mock is somehow self-referential, since `fs-promises.mock.mts` does `export * from "node:fs/promises"`. It isn't: `vi.mock("fs/promises", ...)` in a file that imports nothing else resolves in 196ms. It's the combination that's fatal.

Best guess at the mechanism: mocking a Node builtin stops vitest externalising anything downstream of it, so every module that reaches `fs/promises` has to go through the module runner instead of native import. For this file that's ARCtrl.Javascript, FsSpreadsheet and most of `fable_modules`.

Before this PR the file ran directly through the `output/**/*.test.js` include with no mock, which is the 3.61s path. The new `include` plus `exclude` in `vite.config.mts` routes it through the wrapper instead, so this is new.

`build/Actions/Test.fs:158-175` runs `dotnet fable ... --run npx vitest run` with no timeout of its own, so this likely just sits there.

If the gate only exists for the close-during-load test, it may be cheaper to keep the seam in F# rather than mock a builtin. The test already injects the window via `setBrowserWindowFactory`, so a similar injection point for the load step would avoid `vi.mock` entirely.

tests/Electron.Core/ArcVaultHelper.test.fs
```diff
@@ -259,6 +282,184 @@ let private lifecycleTestWindow id isDestroyed onSend =
...
+        ]
+
+    let setTitle (_value: string) =
+        if destroyed then
+            titleWritesAfterDestroy <- titleWritesAfterDestroy + 1
+            failwith "Object has been destroyed"
+
+    emitJsExpr
+        (windowObject, setTitle)
+        "Object.defineProperty($0, 'title', { configurable: true, get: () => '', set: value => $1(value) })"
```
caroott, line 455: This setter never runs, so `expect(titleWritesAfterDestroy()).toBe(0)` at lines 678 and 704 cannot fail.

`Object.defineProperty` without `enumerable` gives you `enumerable: false`. The mock's `BrowserWindow` constructor (`electron.mock.mts:57-61`) does `Object.assign(this, factory(options))`, and `Object.assign` copies own enumerable properties only. So the instance handed to production code has no `title` property at all, and `this.window.title <- ...` just writes a plain field. I checked in node:

```
'title' in instance: false
descriptor of title on instance: undefined
descriptor of title on source: { get, set, enumerable: false, configurable: true }
titleWritesAfterDestroy = 0   (test asserts 0)
```

The neighbours are fine, for what it's worth: `send` works because `webContents` is a nested enumerable object copied by reference, and `focus` is enumerable too, it just never gets reached on this path.

Delete the `isDestroyed` check in `SetArc`, in `ClearArc`, or the `| Ok _ when this.window.isDestroyed ()` arm in `LoadArc` (ArcVault.fs:388), and this test still passes. For the `LoadArc` one: `SetArc` would not throw against the mock, `RefreshHasUnsavedArcChangesFlag` sends nothing because `false -> false` is not a change, and `Startup`'s own check at line 439 raises the identical message anyway.

Adding `enumerable: true` to the descriptor fixes it.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -123,45 +123,84 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
...
-                    let windowId = windowIdFromIpcEvent event
-                    let! disposition = ARC_VAULTS.OpenOrFocusArc(windowId, arcPath)
-                    return Ok(ArcOpenDisposition.path disposition)
-            with e ->
-                return Error e
+            match! openArcAtPath event arcPath with
+            | Ok disposition -> return Ok(ArcOpenDisposition.path disposition)
+            | Error error -> return Error error
         }
     createARC =
```
caroott, line 228: Not introduced here, but this PR makes it worse.

`createARC` has no `try/with` and no native dialog. If `CreateOrFocusArc` throws, whether from `arc.TryWriteAsyncSwate` failing (ArcVault.fs:492) or `loadWindow` failing (ArcVault.fs:834), the handler rejects. The renderer's `match! Api.ipcArcVaultApi.createARC request` in `Renderer/Components/Helper/ArcVaultHelper.fs:70` then throws rather than matching `Error`, so the `onError` branch right below it never runs.

With `show = false` the window is also invisible now, and `CleanupFailedRegistration` destroys it. So the user picks a folder and gets nothing: no window, no dialog, no toast. Previously a blank window at least appeared.

src/Electron/src/Main/ArcVault/ArcVault.fs
```diff
@@ -705,60 +732,125 @@ type ArcVaults() =
...
+                return Ok()
+            else
+                return
+                    Error(
+                        exn
+                            $"The folder does not contain the required ARC investigation file '{ARCtrl.ArcPathHelper.InvestigationFileName}'."
+                    )
+    }
+
+    member this.RegisterVaultWithArc(path: string) = promise {
```
caroott: Minor: no production caller left after this change. `OpenOrFocusArc` goes straight to the private `RegisterVaultWithValidatedArc`, and the only references to this member are the new tests at lines 619, 658 and 820.

caroott (changes requested):

tests/Electron.Core/ArcVaultHelper.wrapper.test.mts
Etschbeijer: Fixed in commit: 35e0d726afeda0f955cc1d5362b5ff6d6c7e4553

tests/Electron.Core/ArcVaultHelper.test.fs
Etschbeijer: Fixed in commit: af50abd2398f345ed845492dcff24d543c96fd71

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: 459f15463955240ea3b6eab09466c01f41660427

Etschbeijer: Fixed in the commit: 0b4056a0549aca89dd46d518b21d2ea8b2acbc9b

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: 459f15463955240ea3b6eab09466c01f41660427

src/Electron/src/Main/ArcVault/ArcVault.fs
Etschbeijer: Fixed in commit: 684b7eb3211bd126a3a4c17633818ba9df892227

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Implemented in the commit: 1ccfeaf789da73262d559909ab68d2d9be939ff4

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Implemented in the commit: 1ccfeaf789da73262d559909ab68d2d9be939ff4

src/Electron/src/Main/IPC/IArcVaultsApi.fs
```diff
@@ -123,86 +123,151 @@ let private notifyGitRepositoryInitialized (arcPath: string) =
...
+                    with
+                    | Error failure ->
+                        Swate.Components.console.log (
+                            $"Git init failed for '{ArcOpenDisposition.path disposition}': {failure.Message}"
+                        )
+                    | Ok(Some initializedArcPath) -> notifyGitRepositoryInitialized initializedArcPath
+                    | Ok None -> ()
+
+                    return Ok(ArcOpenDisposition.path disposition)
+            with error ->
```
caroott, line 269: `openArcAtPath` lets `ArcLoadCancelledException` through without a dialog, but this `with` turns it into a normal `Error`, so creating an ARC still treats a cancel as a failure.

To hit it: with an ARC open, use "Create a new ARC" from the navbar (that goes down the new-window branch of `CreateOrFocusArc`) and close the new window while it's loading. `Startup` raises the cancellation, cleanup destroys the new window, and the original window shows a "Could not create ARC" modal with an empty body. The body is empty because `exception ArcLoadCancelledException of targetWindowId: int` has no message; Fable compiles it to `super()` and `message` ends up as `""`.

By then the ARC is already written to disk, and the user closed the window on purpose, so there shouldn't be an error at all. Matching the exception here like `openArcAtPath` does is only part of it, though. The renderer's `createArc` calls `onError` for every `Error`, so it also needs a way to tell a cancel apart from a real failure.

tests/Electron.Core/ArcVaultHelper.test.fs
```diff
@@ -917,4 +1998,63 @@ Vitest.describe (
                         Vitest.expect(loadedArc.GetRun("Run With DataMap").DataMap.Value.StaticHash).not.toBe (0)
                     })
         )
+
+        Vitest.test (
+            "OpenARC restores an empty vault after opening a non-ARC folder fails",
```
caroott, line 2003: Nothing checks that a failed open clears `fileTree`. With `this.fileTree.Clear()` removed from `OpenARC`'s `with` branch, all 47 tests in this file still pass. The vault here is fresh, so the tree is empty before and after either way. Seeding the tree before the failed open and asserting it's empty afterwards would cover it.

src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
```diff
@@ -310,17 +310,12 @@ let createWindow () = promise {
             icon = (windowIconPath |> U2.Case2),
             width = int screenSize.width,
             height = int screenSize.height,
+            show = false,
```
caroott, line 313: No test notices if this goes back to `true`. The mock `BrowserWindow` passes `options` to the factory, but none of the factories look at them, so the tests only check that `show()` runs after loading. (Moving `show()` ahead of the load does break three tests, so that part is covered.) Since keeping the window hidden until the renderer has loaded depends on this flag, I'd have the factory record the options and assert `show = false` in one of the registration tests.

tests/Electron.Core/ArcVaultHelper.test.fs
```diff
@@ -917,4 +1998,63 @@ Vitest.describe (
...
+                    Vitest.expect(vault.watcher).toEqual (None)
+                    do! TestHelpers.removeDirectoryAsync folderPath
+                with error ->
+                    do! TestHelpers.removeDirectoryAsync folderPath
+                    return raise error
+            }
+        )
+
+        Vitest.test (
+            "Startup does not start a watcher when ARC loading fails",
```
caroott, line 2028: This test passes even if `Startup` does nothing. `with _ -> ()` swallows any exception, and `watcher = None` is also true when nothing was loaded at all. I replaced `Startup` with an immediate `Promise.resolve()` and this test stayed green. Five other tests did fail, so the suite still catches it, but asserting that `Startup` actually rejected would make this test check what its name says.

caroott (changes requested):

tests/Electron.Core/ArcVaultHelper.test.fs
caroott: The `show` half is covered now. The validation half still isn't: I stubbed both `ValidateArcRoot` calls to return `Ok` in the compiled output and the whole Electron.Core suite stayed green (500/500).

This test passes without the check for two reasons. The loader's own error already mentions `isa.investigation.xlsx`, and `OpenARC`'s rollback resets the vault anyway. So with an empty calling window, the outcome is the same with or without validation.

The check matters when the calling window already has an ARC open. Without it, `OpenOrFocusArc` creates and shows a new window, `OpenARC` fails, and the window gets destroyed again. A test for that path that counts calls to the `setBrowserWindowFactory` handler (expecting zero) would pin it.

src/Electron/src/Main/IPC/IArcVaultsApi.fs
Etschbeijer: Fixed in commits: ee55cf9f6db41a020e93c34d9d70c9a53076dcfd and 2f7d3ffdd06492991ee743f958a60377752652df

tests/Electron.Core/ArcVaultHelper.test.fs
Etschbeijer: Fixed in commit: 1d7cb6c769a20e5596a0ba82ec1562f0f00921b5

src/Electron/src/Main/ArcVault/ArcVaultHelper.fs
Etschbeijer: Fixed in commit: 1d7cb6c769a20e5596a0ba82ec1562f0f00921b5

tests/Electron.Core/ArcVaultHelper.test.fs
Etschbeijer: Fixed in commit:1d7cb6c769a20e5596a0ba82ec1562f0f00921b5

tests/Electron.Core/ArcVaultHelper.test.fs
Etschbeijer: Fixed in commit: 1d7cb6c

